PR #16, opened 2020-03-23
bp: remove dead code

afilini:


dr-orlovsky:
There was some idea that I had to have this code and use it lately, but I forgot what it was the idea :D

I'll try to look at it during my own review of the library and make sense of it; otherwise will delete

dr-orlovsky:
The whole module is removed

Issue #23, opened 2020-04-04
[EPIC]: Add Tor support throughout all protocols


dr-orlovsky:
This will involve modification of rust-bitcoin and rust-lightning libraries and related PRs to them

Upstream PRs:
- [ ] https://github.com/teawithsand/torut/pull/7
- [ ] https://github.com/teawithsand/torut/pull/6
- [ ] https://github.com/teawithsand/torut/pull/5

dr-orlovsky:
Closing since too generic and not related to this repo anymore

PR #24, opened 2020-04-04
Support for Tor addresses

dr-orlovsky:
First step in implementing #23 

The code will be proposed to be adopted in `rust-bitcoin`, or on top of both `rust-bitcoin` & `rust-lightning` as a separate crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,3 +25,7 @@ num-integer = "0.1.42"
 num-traits = "0.2.11"
 num-derive = "0.3.0"
 hex = { version = "=0.3.2" }
+torut = { version = "0.1.1", optional = true }
+
+[features]
+use-tor = ["torut"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,9 @@ num-integer = "0.1.42"
 num-traits = "0.2.11"
 num-derive = "0.3.0"
 hex = { version = "=0.3.2" }
-torut = { version = "0.1.1", optional = true }
+# Move on using "0.1.2" once <https://github.com/teawithsand/torut/pull/1> will
+# be merged and the new version of `torut` crate will be published
+torut = { git = "https://github.com/LNP-BP/torut", branch = "fix-torkey", optional = true }
 
 [features]
-use-tor = ["torut"]
+use-tor = ["torut/v3"]
--- a/src/common/internet.rs
+++ b/src/common/internet.rs
@@ -21,6 +21,13 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use torut::onion::{TorPublicKeyV3, OnionAddressV3, TORV3_PUBLIC_KEY_LENGTH};
 
 
+pub trait UniformAddrEncodable<const L: usize>: Sized {
+    const LEN: usize = L;
+    fn from_uniform_encoding(data: [u8; L]) -> Option<Self>;
+    fn to_uniform_encoding(&self) -> [u8; L];
+}
+
+
 /// A universal address covering IPv4, IPv6 and Tor in a single byte sequence
 /// of 32 bytes.
 ///
@@ -46,13 +53,8 @@ pub enum InetAddr {
     Tor(TorPublicKeyV3),
 }
 
-#[cfg(feature="use-tor")]
-pub const UNIFORM_INETADDR_LEN: usize = TORV3_PUBLIC_KEY_LENGTH;
-#[cfg(not(feature="use-tor"))]
-pub const UNIFORM_INETADDR_LEN: usize = 32;
-
-
 impl InetAddr {
+    #[inline]
     pub fn get_ip6(&self) -> Option<Ipv6Addr> {
         match self {
             InetAddr::IPv4(ipv4_addr) => Some(ipv4_addr.to_ipv6_mapped()),
@@ -62,7 +64,23 @@ impl InetAddr {
         }
     }
 
-    pub fn from_uniform_encoding(data: [u8; UNIFORM_INETADDR_LEN]) -> Option<Self> {
+    #[cfg(not(feature="use-tor"))]
+    #[inline]
+    pub fn is_tor(&self) -> bool { false }
+
+    #[cfg(feature="use-tor")]
+    #[inline]
+    pub fn is_tor(&self) -> bool {
+        if let InetAddr::Tor(_) = self { true } else { false }
+    }
+}
+
+#[cfg(feature="use-tor")]
+const INET_ADD_LEN: usize = TORV3_PUBLIC_KEY_LENGTH;
+#[cfg(not(feature="use-tor"))]
+const INET_ADD_LEN: usize = 32;
+impl UniformAddrEncodable<INET_ADD_LEN> for InetAddr {
+    fn from_uniform_encoding(data: [u8; Self::LEN]) -> Option<Self> {
         match data {
             d if d[0..28] == [0u8; 28] => {
                 let mut a = [0u8; 4];
@@ -81,8 +99,8 @@ impl InetAddr {
         }
     }
 
-    pub fn to_uniform_encoding(&self) -> [u8; UNIFORM_INETADDR_LEN] {
-        let mut buf = [0u8; UNIFORM_INETADDR_LEN];
+    fn to_uniform_encoding(&self) -> [u8; Self::LEN] {
+        let mut buf = [0u8; Self::LEN];
         match self {
             InetAddr::IPv4(ipv4_addr) => buf[24..].copy_from_slice(&ipv4_addr.octets()),
             InetAddr::IPv6(ipv6_addr) => buf[16..].copy_from_slice(&ipv6_addr.octets()),
@@ -94,6 +112,7 @@ impl InetAddr {
 }
 
 impl Default for InetAddr {
+    #[inline]
     fn default() -> Self {
         InetAddr::IPv4(Ipv4Addr::from(0))
     }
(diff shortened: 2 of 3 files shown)


dpc:
`unsafe`? Why not just an enum?

dr-orlovsky:
The idea was to keep all the data within the same byte array, alike the existing `network::Address.address` field. I started using `union` to get some syntaxic sugar over simple [u8; 32], but it ended up with requirement to have `unsafe` code in `union` data access...

dpc:
I don't get it. Is it even going to save any space over an `enum`? Also wouldn't it be easier to to just have one `[u8; 32]` anyway?

I wonder if a different optimization wouldn't be more useful. Since I expect tor addresses too be quite rare, instead of storing them inline, one could store `Box<[u8; 32]>` maybe? So the whole enum is smaller?

dr-orlovsky:
`unoin`s are smaller than `enum`s:  https://doc.rust-lang.org/reference/items/unions.html

However, I have ended up having enum tag anyway (in the parent structure, like in  the sample given in https://doc.rust-lang.org/reference/items/unions.html#pattern-matching-on-unions, so it seems that it is not worth the effort. I will change it just to a simple enum with fixed-length [u8] fields on each case.

dpc:
That was my point: unions are smaller, but only because they don't have the tag. If that tag will still be needed, they will be pretty much the same as enum.

Also, comparing to the tor address size (32B?), the tag is rather insignificant. So if the space usage is important, I would definitely pursue `Box<[u8; 32]>`for tor approach as it slashes 1/2 of the size. 

One could get it all types in exactly 16 bytes, by storing ipv4 as ipv6 with a `::ffff:` prefix([?](https://en.wikipedia.org/wiki/IPv6_address)), and tor as maybe `2001:db8::`([?](https://en.wikipedia.org/wiki/Reserved_IP_addresses#IPv6)) followed by the pointer to boxed tor address. This way some `unsafe` and `union` are still needed, but the external tag is not, and everything fits neatly in 16B, which also have natural alignment.

dr-orlovsky:
Thanks for the detailed considerations. I think that the storage space is actually not important, at least for the cases where rust-bitcoin and rust-lnpbp may be used. Union is appeared as my initial C-background incline, since I saw them natural for the situation — until I went into `unsafe` and other issues, but it was too late. Will update the code to make it w/o union this week and propose to add to `rust-bitcoin`.

Also have found that `rust-ligtning` was also experimenting with universal Tor-supporting addresses https://github.com/rust-bitcoin/rust-lightning/blob/master/lightning/src/ln/msgs.rs#L250, however they intermixed port in there, and I'd prefer to have pure internet addresses with the ability to add port at another layer.

dr-orlovsky:
My follow-up considerations: https://github.com/rust-bitcoin/rust-bitcoin/issues/421#issuecomment-609641715

afilini:
I haven't looked at the code yet, but in my opinion the best thing to do would be to implement it exactly like rust-lightning, or at least that's how I would do it: I don't think it's really worth saving a few bytes using unions, it's not performance critical code. And usually compilers are better than humans at those optimizations :)

dr-orlovsky:
@afilini can you pls check my last considerations from https://github.com/rust-bitcoin/rust-bitcoin/issues/421#issuecomment-609641715 ? And rust-lightning fails to store data if there is only an address, w/o port (i.e. it does not follow Rust `IpAddr/SocketAddr` dichotomy, which is bad and incompatible)

src/common/internet.rs
```diff
@@ -0,0 +1,364 @@
...
+// along with this software.
+// If not, see <https://opensource.org/licenses/MIT>.
+
+///! Universal addresses that support IPv4, IPv6 and Tor
+
+use std::fmt;
+use std::convert::TryFrom;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+#[cfg(feature="use-tor")]
+use torut::onion::{TorPublicKeyV3, OnionAddressV3};
```
afilini, line 20: Why only V3?

Cargo.toml
```diff
@@ -25,3 +25,7 @@ num-integer = "0.1.42"
 num-traits = "0.2.11"
 num-derive = "0.3.0"
 hex = { version = "=0.3.2" }
+torut = { version = "0.1.1", optional = true }
```
afilini, line 28: I would consider specifying features manually since by default it also compiles the controller implementation. I believe `["v2", "v3"]` should be enough: https://github.com/teawithsand/torut/blob/master/Cargo.toml#L17-L18

afilini:
> @afilini can you pls check my last considerations from [rust-bitcoin/rust-bitcoin#421 (comment)](https://github.com/rust-bitcoin/rust-bitcoin/issues/421#issuecomment-609641715) ? And rust-lightning fails to store data if there is only an address, w/o port (i.e. it does not follow Rust `IpAddr/SocketAddr` dichotomy, which is bad and incompatible)

If you want to remove the port that's fine, you can remove it from the enum variants. What i meant is that I think the best way to represent those is with just an enum instead of enum + union for data.

src/common/internet.rs
dr-orlovsky: Because V2 are obsolete already?

Cargo.toml
dr-orlovsky: good point!

dr-orlovsky:
@afilini I don't like rust-lightning design, it not following existing rust `std::net` standards. Anyway, I have spent my time on doing it in line with the standards, w/o `union`s - pls check the last commit in this PR, I have just pushed it.

@dpc just finished the rewrite

src/common/internet.rs
```diff
@@ -0,0 +1,310 @@
...
+    type Error = ();
+    #[inline]
+    fn try_from(addr: InetAddr) -> Result<Self, Self::Error> {
+        Ok(match addr {
+            InetAddr::IPv4(addr) => IpAddr::V4(addr),
+            InetAddr::IPv6(addr) => IpAddr::V6(addr),
+            InetAddr::Tor(addr) => Err(())?,
+        })
+    }
+}
```
afilini, line 110: I would keep `TryFrom` always implemented, even if the `use-tor` feature is disabled: Ideally if i was a user of this library i would prefer to write code that doesn't break if i disable a feature, unless i was explicitly trying to use that feature.

src/common/internet.rs
```diff
@@ -0,0 +1,310 @@
...
+// along with this software.
+// If not, see <https://opensource.org/licenses/MIT>.
+
+///! Universal addresses that support IPv4, IPv6 and Tor
+
+use std::fmt;
+use std::convert::TryFrom;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+#[cfg(feature="use-tor")]
+use torut::onion::{TorPublicKeyV3, OnionAddressV3, TORV3_PUBLIC_KEY_LENGTH};
```
afilini, line 20: Again, probably a stupid question, but why only V3?

src/common/internet.rs
afilini: They are not as good on the privacy level but they are still widely used. I think we should support them if we are trying to do something generic

src/common/internet.rs
```diff
@@ -0,0 +1,310 @@
...
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let mut val = value.to_lowercase();
+        val.push(':');
+        Ok(match &val[..4] {
+            "tcp:" => Transport::TCP,
+            "udp:" => Transport::UDP,
+            "mtcp" => Transport::MTCP,
+            "quic" => Transport::QUIC,
+            _ => Err(())?
+        })
```
afilini, line 263: Not a huge fan of pushing and checking on the longer string. I would just use `starts_with()`

src/common/internet.rs
```diff
@@ -0,0 +1,310 @@
...
+    pub fn get_ip6(&self) -> Option<Ipv6Addr> {
+        match self {
+            InetAddr::IPv4(ipv4_addr) => Some(ipv4_addr.to_ipv6_mapped()),
+            InetAddr::IPv6(ipv6_addr) => Some(*ipv6_addr),
+            #[cfg(feature="use-tor")]
+            _ => None,
+        }
+    }
+
+    pub fn from_32bit_encoding(data: [u8; TORV3_PUBLIC_KEY_LENGTH]) -> Option<Self> {
```
afilini, line 58: "32byte"

src/common/internet.rs
```diff
@@ -0,0 +1,310 @@
...
+        match data {
+            d if d[0..28] == [0u8; 28] => {
+                let mut a = [0u8; 4];
+                a.clone_from_slice(&d[28..]);
+                Some(InetAddr::IPv4(Ipv4Addr::from(a)))
+            },
+            d if d[0..24] == [0u8; 16] => {
```
afilini, line 65: `d[0..16]`

src/common/internet.rs
```diff
@@ -0,0 +1,310 @@
...
+                let mut a = [0u8; 16];
+                a.clone_from_slice(&d[16..]);
+                Some(InetAddr::IPv6(Ipv6Addr::from(a)))
+            },
+            #[cfg(feature="use-tor")]
+            d  => TorPublicKeyV3::from_bytes(&d).map(|pk| InetAddr::Tor(pk)).ok(),
```
afilini, line 71: Not entirely sure about this, but I think you can use `.map(InetAddr::Tor)` without the closure, it should be equivalent.

dr-orlovsky: It is!

src/common/internet.rs
dr-orlovsky: Hm, good point... And if one does not use Tor, they do not need this structures, BTW

src/common/internet.rs
dr-orlovsky: Hm, do they have the same security assumptions?

Basically, by using only V3 in this library we may try to promote usage of V3 and de-usage of V2, if this makes sense...

src/common/internet.rs
dr-orlovsky: Decided to name it in other fashion `from_uniform_encoding`

src/common/internet.rs
```diff
@@ -21,6 +21,13 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use torut::onion::{TorPublicKeyV3, OnionAddressV3, TORV3_PUBLIC_KEY_LENGTH};
 
 
+pub trait UniformAddrEncodable<const L: usize>: Sized {
```
dpc, line 24: As a casual bypasser that is just looking, I can tell you that this could really use a comment explaining what it is for and why does it exist. :)

dr-orlovsky: It does not exist already :) Rust compiler is still very buggy with const generics

Issue #27, opened 2020-04-09
Duplicated type id in Rgb1


afilini:
https://github.com/LNP-BP/rust-lnpbp/blob/ed4fdb314f46e409d1a8f6d93658bfbc352d0d9e/src/rgb/schemata/fungible.rs#L99

https://github.com/LNP-BP/rust-lnpbp/blob/ed4fdb314f46e409d1a8f6d93658bfbc352d0d9e/src/rgb/schemata/fungible.rs#L114

dr-orlovsky:
Thanks for reporting. I see the fix included in #28

dr-orlovsky:
With the introduction of new schema (with pre-defined type ids) the issue become outdated

PR #28, opened 2020-04-09
Validate schema and transitions (closes #7, #27)

### LINKED ISSUE #7: Verification of state transitions against schema

afilini:


src/rgb/schema/transition.rs
```diff
@@ -28,7 +28,7 @@ use crate::csv::{serialize::Commitment, Error};
 #[display_from(Debug)]
 pub struct Transition {
     pub closes: Option<HashMap<usize, Occurences<u32>>>,
-    pub fields: Vec<Field>,
+    pub fields: HashMap<usize, Field>,
```
dr-orlovsky, line 31: I am very worried about custom-sized fields within consensus-critical data. Peter Todd review suggestion was to avoid them at all cost. So we need a fixed size here. BTW, can you pls explain rationale of changing Vec to HashMap?

src/rgb/schema/transition.rs
```diff
@@ -28,7 +28,7 @@ use crate::csv::{serialize::Commitment, Error};
...
     pub binds: HashMap<usize, Occurences<u32>>,
```
dr-orlovsky, line 32: the same as above

src/rgb/schemata/fungible.rs
```diff
@@ -94,36 +102,37 @@ impl Rgb1 {
         //let ts_schema = &schema.transitions[PRIM_ISSUE_TS];
 
         let mut meta = rgb::Metadata::from_inner(vec![
-            metadata::Field { id: metadata::Type(0), val: metadata::Value::Str(String::from(ticker)) },
-            metadata::Field { id: metadata::Type(1), val: metadata::Value::Str(String::from(name)) },
-            metadata::Field { id: metadata::Type(5), val: metadata::Value::U8(precision) },
-            metadata::Field { id: metadata::Type(7), val: metadata::Value::U8(network.into()) },
+            metadata::Field { id: metadata::Type(Self::TICKER_FIELD as u16), val: metadata::Value::Str(String::from(ticker)) },
```
dr-orlovsky, line 105: Probably we just need to change the type of consts to `u16

src/rgb/schemata/fungible.rs
```diff
@@ -139,26 +148,26 @@ impl Schemata for Rgb1 {
                     Self::BALANCE_SEAL => Amount,
                     Self::PRUNE_SEAL => NoState
                 },
-                transitions: vec![
+                transitions: map!{
```
dr-orlovsky, line 151: This was my original code at some point in time, but I moved from the map structure to more straingforward (in serialization) Vecs. I understand that map hives simplier validation code, but my original purpose was to have a DSL-style schema definitions so they may be even written as a standalone file.

I would not insist that this is high priority, but I think it worth preferring readability of the schema over validation code complexity. The reason is the following: validation code is written once, peer-reviewed and checked multiple times; schema will be written many times and most of vendor-specific schemas would not be reviewed. This means we need to give schema vendors tooling to do this not in a error-prone way. We, for sure, can create a language for that (like Ethereum did), but I really think that's a bad idea. So let's just use Rust DLS capabilities <https://doc.rust-lang.org/rust-by-example/macros/dsl.html> to achieve that, avoiding all non-trivial parts (like mixing non-necessary id's)

src/rgb/serialize.rs
```diff
@@ -402,6 +402,7 @@ impl csv::serialize::Commitment for rgb::Transition {
     fn commitment_serialize<E: io::Write>(&self, mut e: E) -> Result<usize, csv::serialize::Error> {
         use crate::rgb::commit::Identifiable;
         Ok(
+            self.id.commitment_serialize(&mut e)? +
```
dr-orlovsky, line 405: Well, this is a commitment serialization; it must have only consensus-critical data. Not sure whether Id is the part of them...

src/rgb/transition.rs
```diff
@@ -15,6 +15,7 @@
 use super::{State, Metadata, Script};
 
 pub struct Transition {
+    pub id: usize,
```
dr-orlovsky, line 18: Same as above: we have to avoid non-statically-sized types (like platform-dependend)

.gitignore
```diff
@@ -17,3 +17,5 @@ Cargo.lock
 Cargo.lock
 
 .idea
+
+*.swp
```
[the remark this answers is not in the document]
afilini: Vim :)

src/rgb/schema/field.rs
```diff
@@ -32,6 +34,35 @@ pub enum FieldFormat {
     Signature(SignatureAlgorithm),
 }
 
+impl FieldFormat {
+    pub fn validate(&self, value: &Value) -> Result<(), ValidationError> {
+        match (self, value) {
```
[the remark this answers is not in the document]
afilini: i personally prefer to keep them this way to avoid mistaking enum variants for structs, but i can adapt if the rest of the code follows different rules. keep in mind that this is still a very very early draft, so many things will change before it's ready to be merged

src/rgb/schema/transition.rs
afilini: what do you mean custom-sized? this is pretty much equivalent to a vec of tuples, with the added bonus that you have indexing on the key.

the idea is that i need to know what the format of a field should be given a "type". before it was just a vec, so let's say that i'm validating the fields and i have a "type 0" that is a string, but multiple string fields defined in the schema. how do match them?

with this when you define the schema you say explicitly that the field of "type 0 should be a string", so when you have a transition that has a field of type 0 you know exactly how to validate it

src/rgb/schemata/fungible.rs
afilini: the problem here is basically the same as above with the fields: you need to match a transition type to the schema when you want to validate

src/rgb/serialize.rs
afilini: it's needed because you don't want to interpret one type of transition as another later. if you don't commit to this value i can change it later

src/rgb/schema/transition.rs
afilini: ah you mean to avoid `usize`? sure, we can do that

src/rgb/schemata/fungible.rs
afilini: the issue was that `HashMap<u16, T>` is not serializable and there's a comment that says that it should be better to only leave the `HashMap<usize, T>` serializable.

So since those constants later are used as keys in the map, they have to be `usize` for the map to be serializable. i was planning to discuss this with you and decide if we want to make other types of maps serializable, or go the other way and cast to usize before using them as keys.

src/rgb/schema/transition.rs
dr-orlovsky: yeah, `usize` thing, it's platform-specific so two apps compiled on different CPUs communicating over network or transferring files to each other will read messaged/data differently

src/rgb/schemata/fungible.rs
dr-orlovsky: A comment? hm... Yeah, let's discuss that and think on it

src/rgb/serialize.rs
dr-orlovsky: Isn't it that the commitment to the serialization order is the same as a commitment to an incremental id?

afilini: the "id" here has one of the possible types of transitions, like "PRIM_ISSUE_TS" or "TRANSFER_TS". in theory if two transitions have similar fields/seals i could create a transition for one of the two types, but then change it later when i validate it. if you commit to the type this is not possible.

i agree that it could be just the index in a vec, but i prefer to have it explicit with the map just to avoid issues for instance if you add another transition and change the order of the others

src/rgb/schema/types.rs
```diff
@@ -14,12 +14,22 @@
 
 use std::{io, convert::TryFrom};
 
-use num_integer::Integer;
 use num_traits::{ToPrimitive, FromPrimitive};
 use num_derive::{ToPrimitive, FromPrimitive};
 
 use crate::csv::serialize::*;
 
+pub trait UnsignedInteger: Clone + Copy + PartialEq + Eq + PartialOrd + Ord + Into<u64> + std::fmt::Debug {
```
dr-orlovsky, line 22: I assume we do not need this as pub? Otherwise it's better to move it into `common` mod.

Cargo.toml
```diff
@@ -23,7 +23,6 @@ grin_secp256k1zkp = { git = "https://github.com/lnp-bp/rust-secp256k1-zkp", bran
 # bitcoin = { version = "0.23.0", features = [ "use-serde" ] }
 rand = "0.5" # Required by grin_secp256k1zkp
 derive_wrapper = "0.1.3"
-num-integer = "0.1.42"
```
dr-orlovsky, line 26: Yeh, the less we have dependencies - the better!

src/rgb/schemata/mod.rs
```diff
@@ -14,10 +14,10 @@
 use super::schema::Schema;
 
 pub mod fungible;
-pub mod collectibles;
+// pub mod collectibles;
```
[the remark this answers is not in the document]
dr-orlovsky:
```suggestion
// TODO: uncomment once RGB-2/collectibles refactoring will be complete
// pub mod collectibles;
```

src/rgb/schemata/mod.rs
```diff
@@ -14,10 +14,10 @@
...
 
 pub use fungible::Rgb1;
-pub use collectibles::Rgb2;
+// pub use collectibles::Rgb2;
```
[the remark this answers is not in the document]
dr-orlovsky:
```suggestion
// TODO: uncomment once RGB-2/collectibles refactoring will be complete
// pub use collectibles::Rgb2;
```

src/rgb/seal.rs
```diff
@@ -48,6 +48,10 @@ impl Seal {
             block_offset: None
         }
     }
+
+    pub fn compare(&self, other: &Self) -> bool {
```
dr-orlovsky, line 52: `ops::Eq`?

dr-orlovsky:
I've added two changes that may affect the code you are working on:
* https://github.com/LNP-BP/rust-lnpbp/pull/39
* https://github.com/LNP-BP/rust-lnpbp/pull/40

If you are ok with them I will merge them into master so you would be able to rebase on it.

PR #33, opened 2020-04-11
Common API modules

dr-orlovsky:


src/api/encode.rs
```diff
@@ -0,0 +1,175 @@
...
+// mutually-exclusive traits (required until negative trait impls will be there)
+// Implemented after concept by Martin Habovštiak <martin.habovstiak@gmail.com>
+mod strategy {
+    use core::marker::PhantomData;
+
+    // Defining strategies:
+    pub enum BitcoinConsensus { }
+    #[cfg(feature="use-rgb")]
+    pub enum RGBStrategy { }
+    pub enum Native { }
```
Kixunil, line 50: It's not clear to me what `Native` means, worth a comment, I think.

src/api/encode.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// 1.2. Auto impl for client-validation-serialized types
+#[cfg(feature="use-rgb")]
+impl<T> MessageEncode for strategy::Holder<T, strategy::RGBStrategy>
+    where T: csv::serialize::Network {
+    type Error = Error;
+    fn into_message(self) -> Message {
+        Message::from(network_serialize(&self.into_inner())
+            .expect("Commitment serialize failed"))
```
Kixunil, line 81: Not strongly related, but serialization failing is quite uncommon thing (unless `serde` is use because of its dynamic nature). Since this is RGB serialization, it'd be nice to try remove the possibility of failure.

src/api/encode.rs
```diff
@@ -0,0 +1,175 @@
...
+
+// 1.1. Auto impl for bitcoin-serialized types
+impl<T> MessageEncode for strategy::Holder<T, strategy::BitcoinConsensus>
+    where T: bitcoin::consensus::encode::Encodable +
+             bitcoin::consensus::encode::Decodable {
+    type Error = Error;
+    fn into_message(self) -> Message {
+        Message::from(consensus_serialize(&self.into_inner()))
+    }
+    fn try_from_message(message: Message) -> Result<Self, Self::Error> {
```
Kixunil, line 69: Forgot to add in the other comment that `&Message` might be more useful as well.

src/api/error.rs
```diff
@@ -0,0 +1,74 @@
...
+use bitcoin;
+use bitcoin::secp256k1;
+
+#[cfg(feature="use-rgb")]
+use crate::csv;
+
+
+#[derive(Debug, Display)]
+#[display_from(Debug)]
+pub enum Error {
```
Kixunil, line 26: I expect the API to be in flux for some time, so making the enum private and bundling it into a newtype struct might improve forward compatibility for the time being.

src/api/encode.rs
dr-orlovsky: In RGB, serialization may fail. After Peter Todd suggestion we have very strict bounds to all possible values; and if some of arrays is going outsized (like it must be `u8` and rust supports only `usize`-based limits for arrays) serialization MAY fail, indicating data integrity problem, that can't be handled at any other level. 

Also, `::network_serialize` is a high-level function, using IO-based write procedure underneath, which is certainly have to be held at upstream level. Even `::network_serialize` itself uses `io::Cursor`, which may fail due to a memory allocation problem, for instance.

Kixunil: Interesting, looks like a pretty good idea. In that case, it might be better to change the signature to `fn to_massage(&self) -> Result<Message, Self::Error>` or  `fn to_massage(&self) -> Result<Message, Self::SerializationError>` to pass the error to the caller. But panicking here is not entirely bad for Rust standards (panic if the problem was caused by the programmer) either. Also notice that I changed it to take `&self` and the name to start with `to` per Rust conventions.

AFAIK `Cursor` doesn't attempt to handle memory allocations by returning `Err`, it's the sub-optimal definition of `Read` and `Write` that force it into returning `io::Result`, but it never returns `Error` anyway. That's pretty much why I wrote [`genio`](https://crates.io/crates/genio).

src/api/encode.rs
dr-orlovsky: If we can use `to_` instead of `into_` for passing references; is there some convention for `from_`s passing references and not consuming the argument?

dr-orlovsky: Also, another important point: we use `impl From` to actually serialize many data structures into Message, and this is just a helper method. If we will pass a reference here, we will have to consume the argument still with the underlying `From`, so I don't understand the reason of changing `self->&self` in `into/from_message` functions here...

src/api/encode.rs
dr-orlovsky: The difference between `network_serialize` and serialization for ZMQ-based APIs/IPCs is the following: network serialization is not an implementation-specific and is a LNPBP standard important to maintain the network "consensus" (all peers can exchange the data in the same way). On the other hand, ZMQ APIs are "vendor-specific"; they are not a part of P2P network consensus and are related either to IPCs between daemons or client-server connections. 

In serialization the error means that the data integrity or code implementation is broken. For P2P networking, it's important to gracefully handle such errors. In APIs, no reason of returning error to the client; we just need to panic and shut down so the developers know that they have to change the code — or the daemon has to restart to do the data integrity checks.

Kixunil: In that case, wouldn't the possibility of an API problem bringing down the whole node be problematic?

src/api/encode.rs
Kixunil: I'm not aware of any convention for from. I've seen `from_*` function take values and references. Just `to_*`/`into_*` are differentiated. I suppose because `Foo::from_something(&bar)` immediately tells the reader what's going on, where `bar.to_foo()` wouldn't if not for the convention.

`From` could be implemented for references too. Implementing it for references only might be even better as it might help avoiding accidental clones.

src/api/error.rs
Kixunil: An advantage of struct over `non_exhaustive` is that it can change its internal representation without breaking backwards compatibility. This may be important in the early stage. Once possible errors are well-understood, making the enum public (possibly with `non_exhaustive`) will be fine.

src/api/encode.rs
dr-orlovsky: It must not be problematic by the design of the node: (1) docker container is always restarted on crash, (2) all operations are transactional and (3) node state integrity is maintained with them and persisted with each state update. Also not the whole node will crash, but a single daemon, like `wired` and this will be detected.

I just can't see how the caller can handle an API serialization error other then crashing as well, so there is no need in propagating it.

Kixunil: OK, thanks for explaining!

src/api/encode.rs
dr-orlovsky: Tried to move from `self`s to `&self`, however got plenty of problems throughout the code; in particular many parts depend on further `into_/from_` consuming `self`s, so I will keep it this way for now and at some point later, when I will have more time, revisit this and refactor the most of codebase

src/api/encode.rs
```diff
@@ -0,0 +1,186 @@
...
+        Ok(Self::new(consensus_deserialize(&message)?))
+    }
+}
+
+// 1.2. Auto impl for client-validation-serialized types
+#[cfg(feature="use-rgb")]
+impl<T> MessageEncode for strategy::Holder<T, strategy::RGBStrategy>
+    where T: csv::serialize::Network {
+    type Error = Error;
```
Kixunil, line 88: The general convention is to not have fixed `Error` type everywhere, but use the original error type. `type Error;` would be unnecessary otherwise, since the trait could require conversion to some specific type. `std::io::{Read, Write}` uses fixed errors (`std::io::Error`) and from my experience, it's quite unfortunate in these cases:

* Application needs to do something smart around errors (real life example: translate messages)
* `no_std` is out of question
* Inconvenient when transferring more information is needed (e.g. both OS code and file name)

So my recommendation is to prefer the original error and let the client of the library unify it if needed.

PR #50, opened 2020-07-08
Various fixes

afilini:
* Fixes again the blinding factors generation
* Changes TxResolver to be a Trait so that we can use a complex struct as a resolver. It's useful if the resolver has some internal state that can be reused, like a connection to an Electrum server or a Bitcoin Core RPC.

### DIFF
--- a/src/rgb/contract/assignments.rs
+++ b/src/rgb/contract/assignments.rs
@@ -59,18 +59,19 @@ impl AssignmentsVariant {
             blinding_factors.push(amount::BlindingFactor::new(&SECP256K1_ZKP, &mut rng));
         }
 
-        // We need the last factor to be equal to the sum
-        let blinding_inputs: Vec<_> = inputs.iter().map(|inp| inp.blinding.clone()).collect();
-        if !blinding_inputs.is_empty() {
-            blinding_factors.pop();
-
-            let blinding_correction = SECP256K1_ZKP
-                .blind_sum(blinding_inputs.clone(), blinding_factors.clone())
-                .expect("SECP256K1_ZKP failure has negligible probability");
-            blinding_factors.push(blinding_correction.clone());
+        // We need the last factor to be equal to the difference
+        let mut blinding_inputs: Vec<_> = inputs.iter().map(|inp| inp.blinding.clone()).collect();
+        if blinding_inputs.is_empty() {
+            blinding_inputs.push(secp256k1zkp::key::ONE_KEY);
         }
 
-        // Composing assignment
+        // remove one output blinding factor and replace it with the correction factor
+        blinding_factors.pop();
+        let blinding_correction = SECP256K1_ZKP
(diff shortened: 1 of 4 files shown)


src/rgb/contract/assignments.rs
```diff
@@ -59,18 +59,19 @@ impl AssignmentsVariant {
...
-            blinding_factors.pop();
-
-            let blinding_correction = SECP256K1_ZKP
-                .blind_sum(blinding_inputs.clone(), blinding_factors.clone())
-                .expect("SECP256K1_ZKP failure has negligible probability");
-            blinding_factors.push(blinding_correction.clone());
+        // We need the last factor to be equal to the difference
+        let mut blinding_inputs: Vec<_> = inputs.iter().map(|inp| inp.blinding.clone()).collect();
+        if blinding_inputs.is_empty() {
+            blinding_inputs.push(secp256k1zkp::key::ONE_KEY);
```
dr-orlovsky, line 65: Here and below: why `ONE_KEY`, not `ZERO_KEY`?

src/rgb/validation.rs
```diff
@@ -205,7 +207,7 @@ pub enum Info {
     UncheckableConfidentialStateData(NodeId, usize),
 }
 
-pub struct Validator<'validator> {
+pub struct Validator<'validator, R: TxResolver> {
```
dr-orlovsky, line 210: This will probably break RGB-node code

dr-orlovsky:
Have posted two questions inline. Also, the build is failing...

src/rgb/contract/assignments.rs
afilini: because i think `ZERO_KEY` is not a valid private key and causes the InvalidSecretKey error sometimes, i believe when there are no other inputs and that's the only blinding factor.

The way I understood this is that the value used as reference is pretty much arbitrary, it just has to be one fixed value that everybody agrees on, but to be honest I'm not a ZK expert or anything, so this should be reviewed by somebody who knows these stuff.

Bottom line is, if you change that to `ZERO_KEY` it crashes, and it doesn't with `ONE_KEY`.

src/rgb/validation.rs
afilini: Yes, but I've already opened lnp-bp/rgb-node#13 that applies the required changes and also adds a proper resolver based on rust-electrum-client

afilini:
The build should be fixed now

src/rgb/validation.rs
dr-orlovsky: sorry, haven't seen it

src/rgb/contract/assignments.rs
dr-orlovsky: Very strange. We will ask Andrew Poelstra to review this as was planned (after test coverage & completion of all code docs), so I will specifically address why this happens.

Also, not happy with the fact that we have to use Grin forked lib for bulletproofs; and will ask him to backport bulletproofs to normal lib.

Issue #60, opened 2020-08-03
Add test coverage for bitcoin-specific strict encodings


dr-orlovsky:
https://github.com/LNP-BP/rust-lnpbp/blob/master/src/bp/strict_encoding.rs

rajarshimaitra:
Working on it.

PR #61, opened 2020-08-05
[WIP]: Add unit tests for Bitcoin specific `strict_encoding`

rajarshimaitra:
This is a draft of unit tests for `strict_encoding` as per #60.

As of now the tests only makes an encode-decode round trip and verifies the integrity of the decoded object. 

Please suggest any form of optimization or missing test cases. 

Will remove [WIP] tag when seems ready for merge.


rajarshimaitra:
Travis is failing at 
```
error: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly
Sometimes not all components are available in any given nightly.
The command "./contrib/test.sh" exited with 1.
```
Which is not occurring at my local and it seem like a rust issue discussed here https://github.com/rust-lang/rustfmt/issues/3404

dr-orlovsky:
Yes, travis failure related to rust itself and it will be fixed only with the next nightly release

Cargo.toml
```diff
@@ -32,7 +32,7 @@ amplify_derive = "~0.1.2"
 lazy_static = "~1.4.0"
 bech32 = "~0.7.2"
 ed25519-dalek = "~1.0.0-pre.3"
-bitcoin = { git = "https://github.com/LNP-BP/rust-bitcoin", branch = "staging" }
+bitcoin = { git = "https://github.com/LNP-BP/rust-bitcoin", branch = "staging", features = ["rand"] }
```
dr-orlovsky, line 35: We will need this only in test configuration, so can you pls move this feature to `dev-dependencies` it as described in https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies

src/bp/strict_encoding.rs
```diff
@@ -374,3 +374,277 @@ impl StrictDecode for bip32::ExtendedPubKey {
...
+
+        // TODO: Network::Other failing the below assertion
+        //let (decoded, written) = encode_decode(&other).unwrap();
+        //assert_eq!(written, 4);
+        //assert_eq!(decoded, other);
+    }
+
+    #[test]
+    fn test_encoding_pubkey() {
+        static PK_BYTES: [u8; 33] = [
```
dr-orlovsky, line 434: I propose to add test vectors with the following options:
- key starting with 0x02
- ZERO_KEY (must be a constant somewhere underneath `secp256k1`)
- ONE_KEY (idem)

src/bp/strict_encoding.rs
```diff
@@ -374,3 +374,277 @@ impl StrictDecode for bip32::ExtendedPubKey {
...
+
+        let (decoded, written) = encode_decode(&ext_pubkey2).unwrap();
+
+        assert_eq!(written, 79);
+        assert_eq!(ext_pubkey2, decoded);
+    }
+
+    #[test]
+    fn test_encoding_script() {
+        let data = Vec::<u8>::from_hex(
```
dr-orlovsky, line 640: Can we pls test more scripts of different type here?

Cargo.toml
rajarshimaitra: Done

src/bp/strict_encoding.rs
rajarshimaitra: ZERO_KEY is removed from secp256k1 library. OKE_KEY test is added.

src/bp/strict_encoding.rs
rajarshimaitra: Added.

rajarshimaitra:
Updated as per the review. This PR is now ready for 2nd look.

dr-orlovsky:
As for garbage, I think PubKey, Signature and OutPoints are enough

rajarshimaitra:
Added garbage tests for `signature`, `pubkey` and `outpiont`. They check for `UnexpectedEof` for now. If that's inadequate and other checks are needed will update accordingly.

rajarshimaitra:
Added `DataIntegrityError` garbage test for pubkey.

PR #62, opened 2020-08-06
Extending RGB PSBT data with helper information

dr-orlovsky:


dr-orlovsky:
Not needed, will be done as a part of standard PSBT key types (extended with tweaking information) at rust-bitcoin level

PR #67, opened 2020-08-25
Test improvements in src/rgb/amount.rs

rajarshimaitra:
Various other test cases including error assertions added. 

Estimated test coverage in src/rgb/amount.rs = 84%

Note:
`Confidential.verify_commit_sum()` is not giving intended result. Added #TODO for future investigation.

### DIFF
--- a/src/rgb/contract/amount.rs
+++ b/src/rgb/contract/amount.rs
@@ -578,11 +578,11 @@ mod test {
 
         let commitments: Vec<secp256k1zkp::pedersen::Commitment> = amounts
             .into_iter()
-            .zip(blinding_factors.into_iter())
+            .zip(blinding_factors.iter())
             .map(|(amount, blinding_factor)| {
                 Revealed {
                     amount,
-                    blinding: blinding_factor,
+                    blinding: blinding_factor.clone(),
                 }
                 .conceal()
                 .commitment
@@ -592,6 +592,31 @@ mod test {
         assert!(Confidential::verify_commit_sum(
             commitments[..positive.len()].to_vec(),
             commitments[positive.len()..].to_vec()
+        ));
+
+        // Create Revealed amounts with wrong positive values
+        let wrong_positive = [1u64, 5u64, 3u64, 4u64, 5u64];
+        let mut amounts = wrong_positive.to_vec();
+        amounts.extend(negative.iter());
+
+        // Create commitments with wrong positive values
+        let wrong_commitments: Vec<secp256k1zkp::pedersen::Commitment> = amounts
+            .into_iter()
+            .zip(blinding_factors.iter())
+            .map(|(amount, blinding_factor)| {
+                Revealed {
+                    amount,
+                    blinding: blinding_factor.clone(),
+                }
+                .conceal()
+                .commitment
+            })
+            .collect();
+
+        // Ensure commit sum verification fails for wrong positive values
+        assert!(!Confidential::verify_commit_sum(
+            wrong_commitments[..positive.len()].to_vec(),
+            wrong_commitments[positive.len()..].to_vec()
         ));
     }
 
--- a/src/rgb/contract/amount.rs
+++ b/src/rgb/contract/amount.rs
@@ -445,9 +445,9 @@ mod serde_impl {
 mod test {
     use super::super::test_helpers::*;
     use super::*;
-    use crate::strict_encoding::StrictDecode;
+    use crate::strict_encoding::{StrictDecode, StrictEncode};
 
-    static AMOUNT: [u8; 43] = [
+    static AMOUNT_65: [u8; 43] = [
         0x3, 0x41, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x20, 0x0, 0xa6, 0x2b, 0x27, 0xae, 0x5a, 0xf,
         0x8c, 0x59, 0x5a, 0xfc, 0x8b, 0x55, 0xe5, 0x5f, 0x72, 0xd7, 0x29, 0x1, 0x55, 0xfa, 0x68,
         0x25, 0xe6, 0x3f, 0x62, 0x73, 0x54, 0xab, 0xfd, 0x11, 0x2e, 0xf5,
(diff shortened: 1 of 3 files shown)


src/rgb/contract/amount.rs
```diff
@@ -445,7 +445,7 @@ mod serde_impl {
 mod test {
     use super::super::test_helpers::*;
     use super::*;
-    use crate::strict_encoding::StrictDecode;
+    use crate::strict_encoding::{StrictDecode, StrictEncode};
 
     static AMOUNT: [u8; 43] = [
```
dr-orlovsky, line 450: Can you pls add comments to which values these amounts are corresponding to?

src/rgb/contract/amount.rs
```diff
@@ -503,15 +503,149 @@ mod test {
...
+        let negetive_commits = negetive
+            .iter()
+            .map(|amount| {
+                Revealed::with_amount(*amount as Amount, &mut rand::thread_rng())
+                    .conceal()
+                    .commitment
+            })
+            .collect::<Vec<secp256k1zkp::pedersen::Commitment>>();
+
+        //TODO: Investigate why commit sum verification fails
```
dr-orlovsky, line 575: It fails since you choose random blinding factors, while they also must add to 0 — see how it's done in https://github.com/LNP-BP/rust-lnpbp/blob/master/src/rgb/contract/assignments.rs#L49 method (and it is the only reason we need that method)

src/rgb/contract/amount.rs
```diff
@@ -503,15 +503,149 @@ mod test {
...
+            negetive_commits
+        ));
+    }
+
+    #[test]
+    fn test_zero_commmit() {
+        let zero_commit = Confidential::zero_pedersen_commitment();
+
+        let mut handmade_bytes = [0x21u8, 0x0u8, 0x08u8].to_vec();
+        handmade_bytes.append(&mut [0x0u8; 32].to_vec());
```
dr-orlovsky, line 587: you can use `extend` and then you would not need `mut` here

src/rgb/contract/amount.rs
```diff
@@ -503,15 +503,149 @@ mod test {
...
+
+        //TODO: Investigate why commit sum verification fails
+        assert!(!Confidential::verify_commit_sum(
+            postive_commits,
+            negetive_commits
+        ));
+    }
+
+    #[test]
+    fn test_zero_commmit() {
```
dr-orlovsky, line 583: Strange, I'm not getting what this method should do and why it works...

dr-orlovsky (changes requested):
Haven't finished the review yet, but figured out why one of the tests fail (pls see comments)

src/rgb/contract/amount.rs
rajarshimaitra: Actually I also forgot what was the value for this amount. I will regenerate the data in the next push.

src/rgb/contract/amount.rs
rajarshimaitra: Hmm, I am still not sure how to do this. Can't seem to find any usage of `verify_commit_sum()` in the above reference.

src/rgb/contract/amount.rs
rajarshimaitra: Yes. better idea.

src/rgb/contract/amount.rs
rajarshimaitra: It first creates a `zero_commit` object from the API. 
Then creates the same object in a serialized format.
Decodes a `commitment` from the serialization.
Checks if the calculated and handmade commitment object matches.

This checks execution of `Confidential::zero_pedersen_commitment()`

dr-orlovsky: I am surprised by the byte value of zero Pedersen commitment though... Doesn't seem right.

src/rgb/contract/amount.rs
dr-orlovsky: Just leave this test as it is; the only thing you should do is to zero-balance blinding factors with secp256k1 function like it is done in the Assignments function from the link above

dr-orlovsky: So, remove the last blinding factor and replace it with the result of the following code:
```rust
       let blinding_correction = SECP256K1_ZKP
            .blind_sum(blinding_inputs.clone(), blinding_factors.clone())
            .expect("SECP256K1_ZKP failure has negligible probability");
        blinding_factors.push(blinding_correction);
```

src/rgb/contract/amount.rs
rajarshimaitra: Yes me to. Theres an weird 0x08 out of nowhere. I almost missed it multiple times when debugging the test. But that's what seems to be the encoding of zero_commit.

src/rgb/contract/amount.rs
rajarshimaitra: Oh, upon trying, we still need `mut` to extend it. Still opting for `extend` as it looks cleaner.

src/rgb/contract/amount.rs
rajarshimaitra: I am trying this. Lets see.

rajarshimaitra:
Pushed with requested changes. I tried balancing the blinding factors. But commit sum verification is still coming out to be false. Blinding factor correction logic seems correct. This is up for a second review.

src/rgb/contract/amount.rs
dr-orlovsky: Good explanation on zero private keys etc is given here: https://crypto.stackexchange.com/questions/49994/find-ecdsa-privkey-to-pubkey-0

So I think we should test private key equal to One, not zero — and test that zero private key results in public key equal to G

dr-orlovsky (changes requested):
Pls apply this diff and the test will work:

```diff
diff --git a/src/rgb/contract/amount.rs b/src/rgb/contract/amount.rs
index 02ed2ee..66bd67b 100644
--- a/src/rgb/contract/amount.rs
+++ b/src/rgb/contract/amount.rs
@@ -552,31 +552,29 @@ mod test {
     #[test]
     fn test_commit_sum() {
         let positive = [1u64, 2u64, 3u64, 4u64, 5u64];
-        let negetive = [7u64, 8u64];
+        let negative = [7u64, 8u64];
 
         // Generate random blinding factors
         let mut rng = rand::thread_rng();
-        let count = positive.len() + negetive.len();
-        let mut blinding_factors = Vec::<_>::with_capacity(count);
+        // We do not need the last one since it is auto-generated to zero-balance the rest
+        let count = positive.len() + negative.len() - 1;
+        let mut blinding_factors = Vec::<_>::with_capacity(count + 1);
         for _ in 0..count {
             blinding_factors.push(BlindingFactor::new(&SECP256K1_ZKP, &mut rng));
         }
 
-        // Adjust Blinding factor for correct summation
-        blinding_factors.pop();
-
         let positive_factors = blinding_factors[..positive.len()].to_vec();
-        let negetive_factors = blinding_factors[positive.len()..].to_vec();
+        let negative_factors = blinding_factors[positive.len()..].to_vec();
 
         let correction = SECP256K1_ZKP
-            .blind_sum(positive_factors, negetive_factors)
+            .blind_sum(positive_factors, negative_factors)
             .unwrap();
 
         blinding_factors.push(correction);
 
         // Create Revealed amounts with corrected blinding factors
         let mut amounts = positive.to_vec();
-        amounts.extend(negetive.iter());
+        amounts.extend(negative.iter());
 
         let commitments: Vec<secp256k1zkp::pedersen::Commitment> = amounts
             .into_iter()
```

dr-orlovsky:
BTW, the original code also worked, so you've done everything right. I just applied a small optimization

rajarshimaitra:
Pushed the above patch. I think it doesn't change the logic of balancing and the test is still failing. What I meant by "failing" is `Confidential::verify_commit_sum()` is giving `false`. I inverted the logic in the test because I didn't want Travis to fail on this unnecessarily. So that probably made the words confusing also because of the way I wrote it. I made the test condition more explicit in this push.

As I was writing the tests for assignments I observed the same behavior in `AssignmentsVariant::zero_balanced()` and `verify_commit_sum()` is giving `false` there too. This confirms the balancing logic of secret keys is probably right here, and something is failing from `SECP256K1_ZKP` library itself.

My proposal:
We should tag this as a potential problem to be investigated and move on. git grepping confirms that we are not using commit sum verification in rgb protocol, though we use zero balancing. So probably important to sort it out, but safe enough for now to move ahead.

dr-orlovsky:
1. Test in my version is not failing on my machine, and it was not negated. You can check my version here: https://github.com/LNP-BP/rust-lnpbp/commit/fa9d4ba1116d464bd5cee40ed0f51c0544aacfbb

2. We can't process w/o this issue solved since `verify_commit_sum` is indeed used in validation logic and is absolutely security-critical:
One - https://github.com/LNP-BP/rust-lnpbp/blob/master/src/rgb/vm/embedded.rs#L108
Two - https://github.com/LNP-BP/rust-lnpbp/blob/master/src/rgb/vm/embedded.rs#L147

3. We need to add wrong sums to make sure that validation fails on wrong sums and runs well on a right sums

rajarshimaitra:
Hmm it seems in your case also the test condition is reversed.
https://github.com/LNP-BP/rust-lnpbp/blob/fa9d4ba1116d464bd5cee40ed0f51c0544aacfbb/src/rgb/contract/amount.rs#L592-L596

Note the `!` in front of `Confidential::verify_commit_sum()`
The test should fail if you remove the `!`.

rajarshimaitra:
Added you changes. Test passing now.

rajarshimaitra:
Pushed changes with failure condition for wrong positive value. Let me know if anything more needed for this PR.

dr-orlovsky:
Can you pls run `cargo fmt`, so Travis would not fail

rajarshimaitra:
I did. That doesn't seem to be working. The latest commit is already cargo formated.

dr-orlovsky:
Right, this is unrelated to the code. Rust again updated the nightly version and forgot to add fmt to the release

PR #68, opened 2020-09-04
Add test coverage to assignments

rajarshimaitra:
This adds test coverage for `AssignmentVariant` structure.
Estimated coverage 78%.

Note

- This depends on [this fix](https://github.com/LNP-BP/rust-lnpbp/pull/67#issuecomment-685867562) to be applied on `AssignmentsVariant::zero_balanced()`. For now the test logic is reversed here.
```
// Check sum verification of commitments
        assert_eq!(
            amount::Confidential::verify_commit_sum(positve, negative),
            false
        );
```
`verify_commit_sum()` should return true once `zero_balanced()` is fixed. Wrong values are not tested here as it has already been covered in test cases for `Ammount` structures.

- This huge structure `PEDERSAN_VARIANT` is this big because of the confidential amounts. Each variant type contains 4 types of assignments, `Revealed`, `Confidential`, `ConfidentialSeal`, `ConfidentialAmount`. As each confidential amount is 712 bytes as encoded data, a pedersan type assignment variant with all 4 types of assignments becomes quite large.

Also let me know if any more test cases to be added.


dr-orlovsky:
If you re-base this onto master it will not require the fix — it's already in the latest commit there.

rajarshimaitra:
Hmm. It seems I am already rebased on top of master. And the "commit_sum_verify()" is still returning false. Probably I am going somewhere off in the testing logic. Will try to debug this tomorrow.

dr-orlovsky:
Right... Here was my fix: https://github.com/LNP-BP/rust-lnpbp/commit/61ce085f955346f189864e9315757c5295630983

Strange. Ok, let's investigate this tomorrow

dr-orlovsky:
Ok, I read the code and see why it's not working. The idea of `zero_balanced` is that it balances inputs and outputs, with outputs split into `ours` (i.e. with known seal data) and `theirs` (only confidential seals). You mix all together: inputs, outputs, ours, theirs (literaly using the same data for both inputs and outputs), and then split them back into inputs and outputs just by filtering with `>3`. This can't work... Pls let me know if I need to provide more explanations on how zero_balanced works

rajarshimaitra:
Updated the `zero_balance()` test to correct logic. Commit sum verification passing now.

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,726 @@ mod strict_encoding {
...
+        let mut bytes = PEDERSAN_VARIANT.clone();
+        bytes[1] = 0x02;
+
+        AssignmentsVariant::strict_decode(&bytes[..]).unwrap();
+    }
+
+    #[test]
+    fn test_zero_balance() {
+        // Base amounts considered as inputs [2, 3, 1, 4]
+        let amounts = [2u64, 3u64, 1u64, 4u64];
```
dr-orlovsky, line 963: Can we use here some more test cases? Like put the main test as a separate fn and run it over multiple cases (including some auto-generated and large ones)?

dr-orlovsky: And it will really help if we originally will separate inputs and outputs, not just by splitting single array into parts

rajarshimaitra: > Can we use here some more test cases?

yes possibly, like creating a function to create arbitrary allocation given a set of input amounts?

>And it will really help if we originally will separate inputs and outputs, not just by splitting single array into parts

Note we are not splitting the array anymore. The entire array is taken as input. Same for the outputs, which is created from the full balanced allocation.

rajarshimaitra:
Added auto-generated testing. It now tests 20 times with randomly generated input vector of random sizes.

dr-orlovsky:
Thank you. While this can be a separate PR, we clearly need to cover the followinng cases:
1. When inputs and outputs are not equivalent in both (a) amounts and (b) count (i.e. sum(inputs) = sum(outputs), but count(inputs) != count (outputs)
2. Edge cases when inputs/outputs contains amounts = `u64::MAX`, `0`
3. Cases where either `ours` or `theirs` is empty
4. Failed tests where sum(inputs) != sum (outputs), including edge situations when they differ for 1 or `u64::MAX`

rajarshimaitra:
Added all the suggested test conditions. Also modified the testing function so that any other conditions can be easily added.

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+            // of random size. Then randomly add 1 to random positions
+            // until the sum of the vector matches with the input sum
+            // This also creates 0 value allocations probabilistically
+
+            let input_sum: u64 = amounts.iter().sum();
+            let mut output_amounts = vec![0u64; rng.gen_range(1, 20)];
+            let output_length = output_amounts.len();
+            for _ in 0..input_sum {
+                output_amounts[rng.gen_range(0, output_length)] += 1u64;
+            }
```
dr-orlovsky, line 1100: With the new value range this will not work efficiently (and will take a lot of time to compute), so I propose to generate `len-1` random numbers, set the last value of either input or output  (depending on which is larger, input or output) to their difference, and the last member of the other one - to zero

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+                &amounts[..],
+                &output_amounts[..],
+                rng.gen_range(0, output_length)
+            ));
+        }
+
+        // Test when input.sum() != output.sum()
+        // When they only differ by 1
+        // When they differ by u64::MAX
+        assert!(!compute_zero_balance(&amounts[1], &amounts[2], 2));
```
dr-orlovsky, line 1113:
```suggestion
        assert!(!compute_zero_balance(&amounts[0], &amounts[1], 2));
        assert!(!compute_zero_balance(&amounts[4], &amounts[5], 2));
        assert!(!compute_zero_balance(&amounts[6], &amounts[7], 2));
        assert!(!compute_zero_balance(&amounts[10], &amounts[11], 2));
```

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+            // until the sum of the vector matches with the input sum
+            // This also creates 0 value allocations probabilistically
+
+            let input_sum: u64 = amounts.iter().sum();
+            let mut output_amounts = vec![0u64; rng.gen_range(1, 20)];
+            let output_length = output_amounts.len();
+            for _ in 0..input_sum {
+                output_amounts[rng.gen_range(0, output_length)] += 1u64;
+            }
+
```
dr-orlovsky, line 1101:
```suggestion
// Check non-equivalent amounts
assert_eq!(compute_zero_balance(&amounts[..(length-1)], &output_amounts[..(output_length-1)], rng.gen_range(0, output_length), false);
```

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+        ));
+
+        // Test random inputs and outputs
+        // Randomly distributed between ours and theirs allocation
+        for _ in 0..10 {
+            // Randomly generate number of amounts between 1 to 20
+            let length = rng.gen_range(1, 20);
+
+            // Randomly fill the amount vector
+            let mut amounts = vec![0; length];
```
dr-orlovsky, line 1081: can we rename `amounts` to `input_amounts` just not to mix with the `amounts` above (which are also used lately aftyer this block)? Also `length` to `input_length`

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+        }
+
+        // Test when input.sum() != output.sum()
+        // When they only differ by 1
+        // When they differ by u64::MAX
+        assert!(!compute_zero_balance(&amounts[1], &amounts[2], 2));
+        assert!(!compute_zero_balance(&[1, 2, 3, 4], &[1, 2, 3, 5], 2));
+        assert!(!compute_zero_balance(
+            &[1, 2, 3, 0],
+            &[1, 2, 3, u64::MAX],
```
dr-orlovsky, line 1117: This test must fail b/c of different reason: amounts can't be larger than `u64::MAX` in their sum, i.e. this line must be equal to 1+2+3-1=5. I think we need to add special test cases to cover overflow situations.

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+            2
+        ));
+    }
```
dr-orlovsky, line 1120: Yes, let's add overflow tests at the very end. For instance:

Test that `u64::MAX + 1` is equal to `0`
Test that `u64::MAX + u64::MAX` is equal to `u64::MAX-1`
Test that `u32::MAX + u32::MAX` is equal to `u64::MAX`
Test that `u32::MAX + u32::MAX + 1` is equal to `0`
Test that `u32::MAX + u32::MAX + 2` is equal to `1`

dr-orlovsky (changes requested):
I am very satisfied with what we are getting with all these tests; it looks like the best test suite among all in this library. Nevertheless. taking a huge importance of this part (basically it's the one that must prevent hidden inflation bugs which may happen with zero knowledge) I have suggested more test cases, which I put as inline suggestions. With that we will be done and I will merge this PR. Very much appreciate your amount of work and thoroughness!

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+            [0u64, 0u64, 0u64, 0u64],
+            [u64::MAX, u64::MAX, u64::MAX, u64::MAX],
+        ];
+
+        for vec in amounts.iter() {
+            assert!(compute_zero_balance(vec, vec, 2));
+        }
+
+        // Test when ours is empty
+        assert!(compute_zero_balance(&amounts[1], &amounts[1], 0));
```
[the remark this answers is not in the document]
rajarshimaitra: Are you sure about this one? This will simply go "index out of bound error". Also it won't test "ours only" which will only happen with `partition=0`.

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+
+        // Test when theirs is empty
+        assert!(compute_zero_balance(
+            &amounts[1],
+            &amounts[1],
+            amounts[1].len()
```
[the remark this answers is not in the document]
rajarshimaitra: I had to break up the `amounts` vector into separate chunks as arrays of different lengths in a single vector is not allowed by the compiler. So I don't really have anything at index 11 here. Can you explain what were you intending to test here?

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+        let amounts = vec![
+            [1u64, 2u64, 3u64, u64::MAX],
+            [10u64, 20u64, 30u64, 40u64],
+            [0u64, 197642u64, u64::MAX, 476543u64],
+            [0u64, 0u64, 0u64, 0u64],
+            [u64::MAX, u64::MAX, u64::MAX, u64::MAX],
+        ];
+
+        for vec in amounts.iter() {
+            assert!(compute_zero_balance(vec, vec, 2));
```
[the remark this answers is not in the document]
rajarshimaitra: Done

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+
+        // test equal inputs and outputs
+        // 4 values equally distributed among ours and theirs allocations
+        // * One zero value case
+        // * All zero value case
+        // * One u64::MAX value case
+        // * All u64::MAX value case
+        // * One zero value + One u64::MAX value case
+        let amounts = vec![
+            [1u64, 2u64, 3u64, u64::MAX],
```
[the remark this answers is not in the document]
rajarshimaitra: Added.

src/rgb/contract/assignments.rs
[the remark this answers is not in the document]
dr-orlovsky: sorry, what I meant was to use `amounts[10]` instead of `amounts[1]`

src/rgb/contract/assignments.rs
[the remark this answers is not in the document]
dr-orlovsky: hm, you should have - I've added new test cases to `amounts` vec above and N1 just shifted to N11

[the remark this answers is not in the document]
dr-orlovsky: It could be I missed with computing the shift, so feel free to update the index. The idea was still to match what there was at N1 index

src/rgb/contract/mod.rs
```diff
@@ -18,9 +18,13 @@ pub mod test_helpers {
     use crate::paradigms::client_side_validation::{CommitEncode, Conceal};
```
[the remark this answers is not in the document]
rajarshimaitra: I tried putting the test helper module at the ned of mod.rs file. But this then gives this error at the macro calls
```
cannot find macro `test_encode` in this scope
have you added the `#[macro_use]` on the module/import?
```
Not sure how exactly to solve this. Putting the module at the top of `mod.rs` file removes this error.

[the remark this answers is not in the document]
dr-orlovsky: Right. Can you move the whole mode into separate file test helpers and just `use` it from the same place in `mod.rs`? This will work (your macro declaration has to be imported before those stuff that uses the macro)

src/rgb/contract/mod.rs
```diff
@@ -63,7 +67,7 @@ pub mod test_helpers {
         );
     }
 
-    pub fn test_confidential<T>(data: &[u8], commitment: &[u8]) -> Result<T, Error>
+    pub(crate) fn test_confidential<T>(data: &[u8], commitment: &[u8]) -> Result<T, Error>
```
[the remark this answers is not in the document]
rajarshimaitra: So far `0x36` doesn't seem to be matching with any enum flag. Do you want this to be changed randomly? In that case, I think there's a fair amount of chance (its an u8) that it will collide with an enum flag.

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+        for _ in 0..10 {
+            // Randomly generate number of amounts between 1 to 20
+            let length = rng.gen_range(1, 20);
+
+            // Randomly fill the amount vector
+            let mut amounts = vec![0; length];
+            for _ in 0..length {
+                // keep the amount value low for faster testing
+                amounts.push(rng.gen_range::<u64>(0, 10));
+            }
```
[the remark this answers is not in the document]
rajarshimaitra: Done

src/rgb/contract/assignments.rs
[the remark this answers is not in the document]
rajarshimaitra: Okay..

src/rgb/contract/assignments.rs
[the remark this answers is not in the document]
rajarshimaitra: Got it.

src/rgb/contract/assignments.rs
```diff
@@ -708,3 +708,813 @@ mod strict_encoding {
...
+            // Create an output amount vector such that
+            // input.sum() = output.sum(), but
+            // input.count() != output.count()
+            // We do this by generating an empty output vector
+            // of random size. Then randomly add 1 to random positions
+            // until the sum of the vector matches with the input sum
+            // This also creates 0 value allocations probabilistically
+
+            let input_sum: u64 = amounts.iter().sum();
+            let mut output_amounts = vec![0u64; rng.gen_range(1, 20)];
```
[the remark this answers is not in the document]
rajarshimaitra: Note: This is not the amount but the length of the output vector.

src/rgb/contract/assignments.rs
rajarshimaitra: Done.

src/rgb/contract/assignments.rs
rajarshimaitra: This [comment](https://github.com/LNP-BP/rust-lnpbp/pull/68#discussion_r485060625). Yes there are many places having very large output vectors will be taxing. This will also create that many confidential amounts which will get huge in memory.

src/rgb/contract/assignments.rs
rajarshimaitra: Hmm i didn't thought about this. Dont we really need to make sure somehow that overflow in confidential amounts doesn't happen? Right now we are getting overflows i think, multiple `u64::MAX` are adding up together and commit sum verification passing.

src/rgb/contract/mod.rs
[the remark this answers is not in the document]
dr-orlovsky: Right, you are changing the first byte, which is the enum. Ok, let's leave it as is

src/rgb/contract/assignments.rs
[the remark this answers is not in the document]
dr-orlovsky: correct, missed that. I meant the value range, not length

src/rgb/contract/assignments.rs
dr-orlovsky: We still need to change the value generation range, and adjusting with 1 would not work for matching the amounts. So this comment remains

src/rgb/contract/assignments.rs
dr-orlovsky: We can't prevent it from happening; however what we can make sure is that the total supply will never be more than `u64::MAX` (which is done at the level of schema), so the sum of all inputs or outputs can't overflow in practice for the assets

dr-orlovsky:
PS If you can add all the requested changes as a new commitment(s) on top of the previous ones, for easier review, that will be very good

src/rgb/contract/assignments.rs
rajarshimaitra: Not sure why it wont work. The value range that we have changed is reflected in the `input_sum`. It will try to push +1 until the input sum matches. This might take a long time. 
Trying to push random values and adjusting the last one might also result into negative value in the end as we don't have control over the random numbers. May be limiting the random numbers to say like 1000 for each value then the remaining into the last item? Something like that?

src/rgb/contract/assignments.rs
rajarshimaitra: I didn't get this one. What are we checking here?

src/rgb/contract/mod.rs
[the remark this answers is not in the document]
rajarshimaitra: Will do that..

src/rgb/contract/assignments.rs
dr-orlovsky: In my first comment I had described an algo which will not result in negative problem

src/rgb/contract/assignments.rs
dr-orlovsky: We are checking non-equivalent amounts: two randomly generated vecs

dr-orlovsky: Since we adjusting the last member of both amount arrays to them will have the same sum, if we will use sub-arrays without the last member, these arrays with extremely large probability will be non-equal in their sum

src/rgb/contract/assignments.rs
dr-orlovsky: What I meant is something like:
```rust
let diff = abs(sum(input) as i128 - sum(output) as i128);
if sum(input) < sum(output) {
   input[input.len()-1) = diff;
} else {
   output[output.len()-1] = diff;
}
```

rajarshimaitra: Ok got it.

src/rgb/contract/assignments.rs
rajarshimaitra: Added

rajarshimaitra:
Added requested updates. Changes are divided into two commits. 1st one refactors the test helpers and 2nd one modifies the assignment tests.
Note:

- The random amounts loop is run for 5 times to keep the total testing time within 60 sec. 

- The overflow conditions has to be modified slightly.

rajarshimaitra:
Travis is complaining on overflows
```
error: this arithmetic operation will overflow
    --> src/rgb/contract/assignments.rs:1181:40
     |
1181 |         assert!(compute_zero_balance(&[u64::MAX + 1], &[0], 1));
     |                                        ^^^^^^^^^^^^ attempt to compute `u64::MAX + 1_u64` which would overflow
     |
     = note: `#[deny(arithmetic_overflow)]` on by default
error: this arithmetic operation will overflow
    --> src/rgb/contract/assignments.rs:1183:15
     |
1183 |             &[u64::MAX + u64::MAX],
     |               ^^^^^^^^^^^^^^^^^^^ attempt to compute `u64::MAX + u64::MAX` which would overflow
error: this arithmetic operation will overflow
    --> src/rgb/contract/assignments.rs:1188:16
     |
1188 |             &[(u32::MAX * u32::MAX - 1) as u64],
     |                ^^^^^^^^^^^^^^^^^^^ attempt to compute `u32::MAX * u32::MAX` which would overflow
error: this arithmetic operation will overflow
    --> src/rgb/contract/assignments.rs:1193:15
     |
1193 |             &[(u32::MAX * u32::MAX) as u64],
     |               ^^^^^^^^^^^^^^^^^^^^^ attempt to compute `u32::MAX * u32::MAX` which would overflow
error: this arithmetic operation will overflow
    --> src/rgb/contract/assignments.rs:1198:16
     |
1198 |             &[(u32::MAX * u32::MAX + 1) as u64],
     |                ^^^^^^^^^^^^^^^^^^^ attempt to compute `u32::MAX * u32::MAX` which would overflow
error: aborting due to 5 previous errors
error: could not compile `lnpbp`.
```

rajarshimaitra:
pushed a fix for Travis.

src/rgb/contract/assignments.rs
```diff
@@ -1039,83 +1038,167 @@ mod test {
...
+            1
+        ));
+        assert!(compute_zero_balance(
+            &[(u32::MAX * u32::MAX) as u64],
+            &[1],
+            1
+        ));
+        assert!(compute_zero_balance(
+            &[(u32::MAX * u32::MAX + 1) as u64],
+            &[2],
```
dr-orlovsky, line 1200:
```suggestion
            &[u64::MAX * u64::MAX + 1],
            &[0],
```

dr-orlovsky (changes requested):

src/rgb/contract/assignments.rs
rajarshimaitra: I am trying to update this. For some reason, the test went really slow and taking over 5 mins. Also not sure why Travis is failing. Will look into these issues tomorrow.

rajarshimaitra:
Pushed the last requested change. `./contrib/test.sh` is passing fully in my local.

rajarshimaitra:
Not sure how to fix this Travis error
```
thread 'rgb::contract::assignments::test::test_zero_balance' panicked at 'attempt to add with overflow', src/rgb/contract/assignments.rs:1182:40
```
Added ` #[allow(arithmetic_overflow)]`, but that doesn't seem to be working.

dr-orlovsky:
I think it is required to use https://doc.rust-lang.org/std/num/struct.Wrapping.html

rajarshimaitra:
As the error is not occurring in my local, can't verify it, but this should work. Let's see if Travis complains.

dr-orlovsky:
We did nearly everything wrong with this overflow tests :) Pls see my fix commitment: https://github.com/LNP-BP/rust-lnpbp/commit/0f8674d39f4eb38e4178ba6f3c33579774183beb

There is no point in overflow testing outside of Pedersen commitment: we test just compiler, not `zero_balance` summation. Then, with Pedersen commitments we operate u256, not u64, so they will not overflow there :)

Also there were two slight issues causing tests to fail non-deterministically, which I fixed.

PR #98, opened 2020-09-15
Test commit encode for Assignments and Metadata

rajarshimaitra:
Commit encoding for `amounts` and `data` is already tested. Commit encoding for `seal` structure is the same as the `conceal` process which is already included in the tests, thus not added here.


dr-orlovsky (changes requested):
Th difference between `commit_encode` and `strinc_encode` procedures is that `comiit_encode` must run "Concel" for it's data and sometime use different strategy of encoding (like Merklization instead of plain encoding). It will be nice to test this parts as well (which require adding more data)

rajarshimaitra:
Merkelization procedure verification added for ancestors. This is preliminary work and seeking for review. If the approach seems ok, will proceed with the rest of the structures.

dr-orlovsky (approved):
I think you did a very good job.

dr-orlovsky:
I had to close the PR b/c I need to release new beta version. So for further changes pls open a new PR starting from the current master state

Issue #100, opened 2020-09-16
Move network field from Schema to Genesis


dr-orlovsky:
When we have network in Schema, this will result in that the same schema will have a different SchemaId on different networks (mainnet, testnet), which is undesirable. In fact, network is an asset-specific thing (i.e. genesis is more natural place), and not smart contract category-specific.

Next, Genesis must include full ChainParams data structure, not just a network ID. This will allow RGB to operate on top of any future chain (like new federated sidechains) without requiring library updates.

Related issues: 
- https://github.com/LNP-BP/rust-lnpbp/issues/102 Refactor bp::Network into ChainParams
- https://github.com/LNP-BP/rust-lnpbp/issues/95 Use ChainParams instead of Network

Issue #95, opened 2020-09-14
RGB: Use ChainParams instead of bp::Network


dr-orlovsky:
Upstream dependencies:
- [ ] https://github.com/rust-bitcoin/rust-bitcoin/issues/481
- [ ] https://github.com/LNP-BP/rust-lnpbp/issues/102

PR #104, opened 2020-09-22
Refactoring bp::Network into Chain and ChainParams: closes #102

### LINKED ISSUE #102: Refactor bp::Network into ChainParams
We need to support wide range of networks (including Liquid, LN) and create a foundation for generic asset concept (that includes chain-based assets and RGB assets).

Draft concept:
```rust

/// P2P network magic number: prefix identifying network on which node operates
pub type P2pMagic = u32;
/// Magic number prefixing Pubkey or Prvkey data according to BIP32 spec
pub type Bip32Magic = u32;

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
#[non_exhaustive]
#[repr(u8)]
pub enum ChainFormat {
    Bitcoin = 0,
    Elements = 1,
}

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
#[repr(u8)]
pub enum AssetLayer {
    Layer1and2,
    Layer2and3,
}

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct AssetParams {
    pub ticker: &'static str,
    pub unit_of_accounting: &'static str,
    pub indivisible_unit: &'static str,
    pub divisibility: u64,
    pub asset_id: AssetId,
}

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct ChainParams {
    pub name: &'static str,
    pub p2p_magic: P2pMagic,
    pub genesis_hash: BlockHash,
    pub bip32_pubkey: Bip32Magic,
    pub bip32_prvkey: Bip32Magic,
    pub bip70_name: &'static str,
    pub bip173_prefix: &'static str,
    pub p2p_port: u16,
    pub rpc_port: u16,
    pub ln_height: u32,
    pub format: ChainFormat,
    pub dust_limit: u64,
    pub native_asset: AssetParams,
    pub testnet: bool,
    pub pow: bool,
}

/// A set of recommended standard networks. Differs from bitcoin::Network in
/// ability to support non-standard and non-predefined networks
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
#[non_exhaustive]
#[repr(u32)]
pub enum Chains {
    Mainnet,
    Testnet3,
    Regtest(BlockHash),
    Signet,
    SignetPrivate(BlockHash),
    LiquidV1,
    Other(ChainParams),
}

```

dr-orlovsky:
Lightning network supports multiple chains, so bitcoin mainnet, testnet, signet, liquid and even non-bitcoin litecoin are part of the same network. Because of this LN uses chain hash - and LN nodes internally define a set of chain parameters to operate with different networks. In order to be interoperable with this concept, and in order to increase Liquid compatibility this PR implements generic concept of Chains and ChainParams as a part of `bp` mod (bitcoin protocol) and moves RGB from using just a P2P network magic bytes in Schema to encoding complete chain parameter data as a part of asset genesis (restricting commitment only to the chain genesis block hash). See attached issues for more details: #102, #100 & #95, as well as https://github.com/LNP-BP/LNPBPs/issues/58


src/bp/chain.rs
```diff
@@ -0,0 +1,1413 @@
...
+    }
+}
+
+hash_newtype!(
+    AssetId,
+    sha256d::Hash,
+    32,
+    doc = "Universal asset identifier for on-chain and off-chain assets; for \
+           on-chain assets matches genesis hash of the chain, but displayed in \
+           normal, non-reverse order",
```
rajarshimaitra, line 171: I am a little confused by the wording here. Will the `AssetId` **match** with genesis chain hash? Or it will **commit** to the genesis chain hash?

rajarshimaitra (commented):
Overall code looks good to me. This will increase flexibility by a lot. 
Just one small question below.

src/bp/chain.rs
```diff
@@ -0,0 +1,1413 @@
...
+        ))
+    }
+}
+
+/// A set of recommended standard networks. Differs from bitcoin::Network in
+/// ability to support non-standard and non-predefined networks
+#[derive(Clone, Debug, Hash)]
+#[non_exhaustive]
+#[repr(u32)]
+pub enum Chains {
```
rajarshimaitra, line 593: Will it be better to call this `Chain` instead of `Chains`, that might give the false impression that an asset can have multiple chains.

dr-orlovsky: You are right, I need to change the name

PR #112, opened 2020-09-30
RGB "decentralized geneses": contract unowned extensions & valencies implementation

dr-orlovsky:
See https://github.com/LNP-BP/LNPBPs/issues/52 for explanations. This PR implements:
- [x] Schema for extensions
- [x] Schema validation
- [x] Extension implementation within RGB contracts
- [x] Inclusion of extension data into consignments
- [x] Consignments validation for extensions

@rajarshimaitra can you pls update test data to make tests not failing? And we need to add test coverage where it is missed.


rajarshimaitra:
Noted, will work on the test updates.

rajarshimaitra:
Yeah. Finally managed to push it.

dr-orlovsky:
@rajarshimaitra can you also pls work on adding cases to existing tests here so we would not forget it later? (I.e. don't write new tests; just add coverage for new fields when needed if the test for other fields/structures are already in place)

PR #113, opened 2020-09-30
Refactoring repository into multiple crates

dr-orlovsky:
Required for using derive macros, which depend on one part of the core library in other parts of the core library


dr-orlovsky:
Actually, refactoring is not required: I found a way how I can import derive macros into the main crate

PR #119, opened 2020-10-05
Improvements in garbage testing macro

rajarshimaitra:
Updated the garbage testing macro. Though it cant be generalized for all cases. Edited some previous garbage macro to explicit form, which cant be covered with the new garbage testing macro.

### DIFF
--- a/src/paradigms/strict_encoding.rs
+++ b/src/paradigms/strict_encoding.rs
@@ -1012,7 +1012,7 @@ pub mod test {
         } };
     }
 
-    // Macro to run test_suite
+    /// Macro to run test_suite
     #[macro_export]
     macro_rules! test_encode {
         ($(($x:ident, $ty:ty)),*) => (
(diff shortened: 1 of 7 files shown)


dr-orlovsky:
Have you tried making macro covering multiple enum ranges, with the suggestion we discussed? (I was not sure if rust compiler is able to do that though)
```rust
macro_rules! test_garbage_exhaustive {
    ($($range:expr)+; $( ($x:ident, $ty:ty), )+ ) => (
        {$(
            let mut cp = $x.clone();
            $( for byte in $range {
                cp[0] = byte as u8;
                assert_eq!(
                    <$ty>::strict_decode(&cp[..]).unwrap_err(),
                    Error::EnumValueNotKnown(stringify!($ty), $byte)
                );
            })+
        )+}
    );
}
```

dr-orlovsky (approved):

rajarshimaitra:
Yes, I started with this initially. But it was giving some pattern matching error in the macro. So modified it to only having a single range. I think it's not possible to have repetition in ranges and the objects independently. To have multiple ranges, we can include the range inside the tuple as (range, Object, Type), and test the same Object for different ranges. 
For most of the cases though, as Enums are only a byte, its enough to have a single range for all the enums. Thus changed it into single range pattern.

dr-orlovsky:
Thanks, I had doubts that it will work... I probably had few disjoint ranges for some enums (like with bit flags), but its really rare...

dr-orlovsky:
Occasionally closed instead of pressing merge

PR #121, opened 2020-10-05
Added test coverage for schema::state.rs

rajarshimaitra:
This adds basic testing coverage for `schema::state.rs`. Coverage is satisfactory. I tried to touch all the parts of the module. Let me know if anything more needs to be added.

Note: 
There are two possible issues worth future investigation. I have marked them with `TODO`. 
The first one `Investigate why ::core::mem::size_of_val(&min) is not giving correct length` is probably a bug. I modified the encoding logic to pass the tests for now, which might not be the best approach to do it. Open to other suggestions.

### DIFF
--- a/src/rgb/schema/state.rs
+++ b/src/rgb/schema/state.rs
@@ -316,8 +316,8 @@ mod strict_encoding {
                     {
                         Err(Error::DataIntegrityError(format!(
                             "Lower bound {:?} of the allowed range for \
-                                 DataFormat is outside of the possible values \
-                                 of used number type",
+                             DataFormat is outside of the possible values \
+                             of used number type",
                             bound,
                         )))?
                     }
@@ -325,7 +325,7 @@ mod strict_encoding {
                     Bound::Excluded(_) if !exclusive => {
                         Err(Error::DataIntegrityError(
                             "Excluded upper bound for the allowed range in \
-                         DataFormat does not make sense for float type"
+                             DataFormat does not make sense for float type"
                                 .to_string(),
                         ))?
                     }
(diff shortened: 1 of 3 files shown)


dr-orlovsky:
Can you pls uncomment/unfix all tests under question so they will fail and I can debug?

src/rgb/schema/state.rs
```diff
@@ -1055,3 +1060,582 @@ mod _validation {
...
+        assert_eq!(
+            enum_fromat.validate(3, &Revealed::U16(4u16)).failures[0],
+            Failure::SchemaMismatchedBits {
+                field_or_state_type: 3,
+                expected: Bits::Bit8
+            }
+        );
+
+        // Test failure cases for String format
+        // TODO: Add failure cases when provided size is less the permitted
```
dr-orlovsky, line 1444: The test case seem to be present, is this an outdated TODO comment?

src/rgb/schema/state.rs
```diff
@@ -1055,3 +1060,582 @@ mod _validation {
...
+            string_format.validate(3, &string_data).failures[0],
+            Failure::SchemaWrongDataLength {
+                field_or_state_type: 3,
+                max_expected: 2,
+                found: 5
+            }
+        );
+
+        // Test failure cases for Bytes format
+        // TODO: Add failure cases when provided size is less the permitted
```
dr-orlovsky, line 1458: Same as above

src/rgb/schema/state.rs
```diff
@@ -363,10 +363,15 @@ mod strict_encoding {
...
+                    let (min, max) = (
+                        $min.to_le_bytes().to_vec(),
+                        $max.to_le_bytes().to_vec(),
+                    );
                     $e.write_all(&min)?;
                     $e.write_all(&max)?;
-                    $len += ::core::mem::size_of_val(&min) * 2
+                    // TODO: Investigate why ::core::mem::size_of_val(&min)
+                    //       is not giving correct length
+                    $len += min.len() * 2
```
dr-orlovsky, line 374: Of course this is a bug, I had to write `&$min`, not `&min` (so it computes length of a sliced type, which in contains length field in itself, not of a primitive type). Fixed

dr-orlovsky (approved):
Fixed issues in a separate commit, doing merge

PR #123, opened 2020-10-06
Added test coverage for schema::nodes

rajarshimaitra:
Basic test coverage for schema::nodes. 

Added a variant in `GensisAction` and `ExtensionAction` to create the schemas. Hope it won't be a big problem.

### DIFF
--- a/src/rgb/schema/nodes.rs
+++ b/src/rgb/schema/nodes.rs
@@ -496,7 +496,7 @@ mod test {
     }
 
     #[test]
-    fn test_nodetraits_for_genesis() {
+    fn test_node_for_genesis() {
         let genesis_schema =
             GenesisSchema::strict_decode(&GENESIS_SCHEMA[..]).unwrap();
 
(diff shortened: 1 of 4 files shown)


src/rgb/schema/nodes.rs
```diff
@@ -205,12 +205,14 @@ mod strict_encoding {
             &self,
             mut e: E,
         ) -> Result<usize, Error> {
-            self.metadata.strict_encode(&mut e)?;
-            self.defines.strict_encode(&mut e)?;
-            self.valencies.strict_encode(&mut e)?;
-            self.abi.strict_encode(&mut e)?;
+            let mut len = 0usize;
+            len += self.metadata.strict_encode(&mut e)?;
```
rajarshimaitra, line 209: These are added to return the encoded length correctly.

dr-orlovsky (commented):
I did all nits as a separate commit; the rest is LGTM

Issue #134, opened 2020-10-29
Implement Noise encryptor/decryptor (transcoder)


dr-orlovsky:
We need it in this module, with exactly the same API like `NoEncryption` struct: https://github.com/LNP-BP/rust-lnpbp/blob/master/src/lnp/session/transcoders.rs#L127-L186

It must be splittable with `Bipolar` trait into encryptor and decryptor separate components (required for multi-thread nonblocking TCP workflow).

It must be implemented in form of a handshake state machine, alike it was made before when I was using rust-lightning noise encoder: https://github.com/LNP-BP/rust-lnpbp/blob/v0.1.0-rc.1/src/lnp/session/connection.rs#L73-L174 (but unlike that code no multithreading/TCP management is required). Basically you need to keep the internal state and "eat" initial three messages representing handshake, switching internal state. Another sample is state machine noise refactoring from the rust-lightning: https://github.com/rust-bitcoin/rust-lightning/pull/691

PR #135, opened 2020-10-31
Implement Noise Handshake (BOLT8)

rajarshimaitra:
This implements the Noise handshake protocol as per BOLT8 as per [Issue134](https://github.com/LNP-BP/rust-lnpbp/issues/134). This PR mostly uses the BOLT8 handshake structures implemented [here](https://github.com/rust-bitcoin/rust-lightning/pull/691), which is refactored to suit our needs. 

Following are brief descriptions of modules in `src/lnp/session/noise/` directory

`act`: This describes the handshake action structures as defined in BOLT8.\
`chacha`: Encryption and Decryption functions for ChaCha20Poly1305 AEAD encryption.
`hkdf`: Key derivation functions as defined in BOLT8
`conduit`: Connection conduit that has read buffer and implements automatic key rotation as per BOLT8. This structure is also implemented with `Transcode` and `Bipolar` to make the API compatible with other parts of the library.
`handshake`: The Handshake states and their transition logic as per BOLT8.

This is more or less self contained at this moment. The Noise module is private and none of the functionalities are used elsewhere.

A previous `noise.rs` file which had basic AEAD examples has been removed. 

Test cases are included to indicate the proper functioning of the module.

Open to suggestion on how to properly credit the above PR author in rust-lightning.


src/lnp/session/noise/act.rs
```diff
@@ -0,0 +1,178 @@
+// LNP/BP Core Library implementing LNPBP specifications & standards
+// Written in 2020 by
+//     Dr. Maxim Orlovsky <orlovsky@pandoracore.com>
```
dr-orlovsky, line 3: I think the credits should go here, instead of my name. We need to place (1) authors of Rust lightning original handshake code (Matt Corallo/Antonie Riard), (2) guys that proposed PR and youself. So probably:

```suggestion
//     Julian Knutsen
//     Rajarshi Maitra
//     Arik Sosman
//     Matt Corallo
//     Antoine Riard
```

src/lnp/session/noise/chacha.rs
```diff
@@ -0,0 +1,129 @@
+// LNP/BP Core Library implementing LNPBP specifications & standards
+// Written in 2020 by
+//     Dr. Maxim Orlovsky <orlovsky@pandoracore.com>
```
dr-orlovsky, line 3: Again, just replace my name with names of those who actually written the code

src/lnp/session/noise/conduit.rs
```diff
@@ -0,0 +1,630 @@
...
+use super::handshake::HandshakeError;
+use chacha20poly1305::aead::Error;
+use std::borrow::Borrow;
+
+pub type SymmetricKey = [u8; 32];
+
+/// Maximum Lightning message data length according to
+/// [BOLT-8](https://github.com/lightningnetwork/lightning-rfc/blob/v1.0/08-transport.md#lightning-message-specification)
+/// and [BOLT-1](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#lightning-message-format):
+const LN_MAX_MSG_LEN: usize = ::std::u16::MAX as usize; // Must be equal to 65535
```
dr-orlovsky, line 14: We already have this constant here: https://github.com/LNP-BP/rust-lnpbp/blob/master/src/lnp/transport/mod.rs#L34-L46

It's better to use them, not create a new one

src/lnp/session/noise/conduit.rs
```diff
@@ -0,0 +1,630 @@
...
+pub struct Encryptor {
+    sending_key: SymmetricKey,
+    sending_chaining_key: SymmetricKey,
+    sending_nonce: u32,
+}
+
+impl Encryptor {
+    pub fn encrypt(&mut self, buffer: &[u8]) -> Result<Vec<u8>, Error> {
+        if buffer.len() > LN_MAX_MSG_LEN {
+            panic!("Attempted to encrypt message longer than 65535 bytes!");
```
dr-orlovsky, line 31: No panic should happen here but error reporting. I know that error would not propagate, but we don't need it to propagate: we have message size check at the transport level which happens before the encryption: https://github.com/LNP-BP/rust-lnpbp/blob/master/src/lnp/transport/ftcp.rs#L158

src/lnp/session/noise/conduit.rs
```diff
@@ -0,0 +1,630 @@
...
+    }
+}
+
+impl Encrypt for Encryptor {
+    fn encrypt(&mut self, buffer: impl Borrow<[u8]>) -> Vec<u8> {
+        match self.encrypt(buffer.borrow()) {
+            Ok(values) => return values,
+            Err(_) => return Vec::new(),
+        }
+    }
```
dr-orlovsky, line 80: Later I will refactor `Encrypt` trait such that it may return error. But for now this will work

src/lnp/session/noise/conduit.rs
```diff
@@ -0,0 +1,630 @@
...
+            );
+        }
+
+        let (current_message, offset) = self.decrypt(&read_buffer[..])?;
+        read_buffer.drain(..offset); // drain the read buffer
+        self.read_buffer = Some(read_buffer); // assign the new value to the built-in buffer
+        Ok(current_message)
+    }
+
+    fn decrypt(
```
dr-orlovsky, line 131: Not sure whether it's nice idea to have the same function name in the impl of the type as in the trait. Probably `decrypt_raw` here will work better

src/lnp/session/noise/conduit.rs
```diff
@@ -0,0 +1,630 @@
...
+        }
+    }
+}
+
+/// Returned after a successful handshake to encrypt and decrypt communication
+/// with peer nodes. It should not normally be manually instantiated.
+/// Automatically handles key rotation.
+/// For decryption, it is recommended to call `decrypt_message_stream` for
+/// automatic buffering.
+pub struct Conduit {
```
dr-orlovsky, line 249: Many be name it as `NoiseEncryption` or `LnEncryption`? To match `NoEncryption` name?

dr-orlovsky (approved):
utACK. Only a few nits, otherwise looks great.

I will check locally that it does not break master (which I refactored yesterday in node address types) and will merge after

dr-orlovsky (approved):

src/lnp/session/noise/conduit.rs
rajarshimaitra: Oh, it seems this was a bug. the `ftcp` msg size check is as below
```
if len > super::MAX_FRAME_SIZE {
            return Err(Error::OversizedFrame(len));
        }
```
where `MAX_FRAME_SIZE = 0xFFFF + 34`. But in `conduit` above the encryption length is being checked for `LN_MAX_MSG_LEN` which is `0xFFFF`. So even if the size was correct as per `ftcp` check, it would have panicked here.

My suggestion is we get rid of the check here and add documentation warning users that they should check for max msg length before passing for encryption here.

rajarshimaitra: Also, should we use `MAX_FRAME_SIZE` which is `0xFFFF + 34` in all places which have check against `LN_MAX_MSG_LEN`? It should depend on the context but my question is, in general, we should check for the maximum allowed frame size which is `0xFFFF + 34` instead of the maximum allowed msg size which is `0xFFFF`, right?

src/lnp/session/noise/conduit.rs
rajarshimaitra: Ya noted the same. I thought of doing the refactoring myself to return `Result` but I thought it would need modification elsewhere so would be better to do it in separate PR.

src/lnp/session/noise/conduit.rs
rajarshimaitra: Noted.

src/lnp/session/noise/conduit.rs
rajarshimaitra: Noted. Going for `NoiseEncryption`.

src/lnp/session/noise/conduit.rs
rajarshimaitra: I am juggling between `MAX_FRAME_PAYLOAD_SIZE` and `MAX_FRAME_SIZE`. It seems `MAX_FRAME_SIZE` would be a more appropriate constant to use here.

src/lnp/session/noise/act.rs
rajarshimaitra: Done

src/lnp/session/noise/chacha.rs
rajarshimaitra: Done

dr-orlovsky:
Just would like to know to thank @julianknutsen and let him know that we used the code he wrote for rust-lightning (but was never accepted) in our rust Lighting Network Protocol implementation (which is otherwise independent from the rust-lightning).

This implementation is used in lightning node https://github.com/LNP-BP/lnp-node by LNP/BP Standards association, designed specially for supporting such projects as RGB, DLCs and future LN extensions (eltoo, PTLCs etc).

CC @Tibo-lg

julianknutsen:
I appreciate the call out and good luck with the project. I'm glad the code found a good use case here and that it was able to help the larger LDK community.

src/lnp/session/noise/conduit.rs
dr-orlovsky: > My suggestion is we get rid of the check here and add documentation warning users that they should check for max msg length before passing for encryption here.

No need, I think. While handshake is part of BOLT-8, it's in the other part and related to LN message encoding but not to a specific encryption scheme. I am pretty much sure that this noise encryption can be used in other non-LN cases which do not require message size limit

PR #137, opened 2020-11-02
Some basic housekeeping in noise handshake

rajarshimaitra:
Made some basic housekeeping changes I found while taking a more detailed look.

Tried a new thing. Added comments at change locations, with tag `PR Comment` with an explanation of the change. Which can be easily removed while merging, but makes review easier, especially when code is removed.

Let me know if it's helpful at all, or I will remove it.


dr-orlovsky (approved):
LGTM

PR #139, opened 2020-11-02
add macos to the matrix strategy

dieeasy:
* don't fail fast so a failing arch doesn't cancel the remaining jobs
 * install dependencies based on matrix.os
 * uniform ${{ }} syntax usage


dr-orlovsky (approved):
LGTM!

PR #141, opened 2020-11-06
CI: add android + ios builds

dieeasy:
some points that might require special attention during review:
- `zeromq-src` from https://github.com/LNP-BP/zeromq-src-rs is now applied globally instead of just for android
- `lightning-invoice` is now fetched from https://github.com/LNP-BP/rust-lightning-invoice to avoid the same error as in https://github.com/LNP-BP/rgb-node/issues/4
- `openssl` is now vendored for ios
- these new jobs use `cargo check` instead of `cargo build`, as in `467ba2a80e945c3e4915a43b3ee82bde08b6607b`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-18.04, ubuntu-20.04, macos-10.15, macos-11.0 ]
+        os: [ ubuntu-18.04, ubuntu-20.04, macos-10.15, macos-11.0, windows-2019 ]
     steps:
       - uses: actions/checkout@v2
       - name: Install linux dependencies
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -253,3 +253,22 @@ jobs:
           export CXXFLAGS="$CFLAGS -nostdlib++ -I$NDK_HOME/sources/cxx-stl/llvm-libc++/include"
           export LDFLAGS="--sysroot=$NDK_HOME/platforms/android-21/arch-x86"
           cargo check --all-features --target=i686-linux-android
+  ios:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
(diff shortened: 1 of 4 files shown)


dr-orlovsky:
> zeromq-src from https://github.com/LNP-BP/zeromq-src-rs is now applied globally instead of just for android

This does not work that way. The patch-io section only works for the current crate and do not propagate to the crates using it. Meaning that when you add it here the downstream dependency will still compile `lnpbp` crate against unpatched version! the only way of patching which zmq version `lnpbp` crate use is to patch it at the level of `rgb-sdk` directly (and the patch will apply to all dependencies there)

> lightning-invoice is now fetched from https://github.com/LNP-BP/rust-lightning-invoice to avoid the same error as in LNP-BP/rgb-node#4

Again, we can't do it since this will prevent this crate from being published/release (we can only publish/release crates with dependencies on published crates). The only way is to apply the patch at the topmost level (i.e. rgb-sdk)

> openssl is now vendored for ios

agree

> these new jobs use cargo check instead of cargo build

I think I already fixed it in https://github.com/LNP-BP/rust-lnpbp/commit/467ba2a80e945c3e4915a43b3ee82bde08b6607b

dieeasy:
- dropped `zeromq-src`
- restored published `lightning-invoice` version
- rebased on `master`
- removed some unnecessary steps and components (faster now)
- updated `cmdline-tools` and added sha256sum check on the downloaded zip file

also, dropped 32-bit android architectures and ios for the moment, as they fail to build now and can be added later

I think this is now ready for you to review again

---
>> these new jobs use cargo check instead of cargo build, as in 467ba2a80e945c3e4915a43b3ee82bde08b6607b

> I think I already fixed it in 467ba2a

yep, meant to say that I noticed the change you introduced and updated accordingly

dieeasy:
going back to `lightning-invoice` and the fact that it would prevent this crate from being published, I can confirm (used `cargo publish --dry-run` as reference) that it fails when replacing the current dependency but succeeds when patching it

I'm thus going to update this PR patching lightning-invoice and re-introducing android 32-bit targets

dieeasy:
ios and windows have been added, this should now be complete

dr-orlovsky (approved):
Thank you very much!

PR #144, opened 2020-11-10
BigSize (BOLT 1) implementation

rajarshimaitra:
A BigSize implementation with some nits around the code base. 

Encoding Decoding is done using `Encode/Decode` trait from `src/lnp/presentation/encoding`. If any other form of encoding is required that can be updated.

Test cases from BOLT1 is added for BigSize in `mod.rs` itself.

### DIFF
--- a/src/lnp/presentation/error.rs
+++ b/src/lnp/presentation/error.rs
@@ -41,18 +41,18 @@ pub enum Error {
     /// unknown LNP protocol version
     UnknownProtocolVersion,
 
-    /// error in strict encoded data in LMP message
+    /// error in strict encoded data in LNP message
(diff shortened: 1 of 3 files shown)


src/lnp/presentation/error.rs
```diff
@@ -41,18 +41,18 @@ pub enum Error {
     /// unknown LNP protocol version
     UnknownProtocolVersion,
 
-    /// error in strict encoded data in LMP message
+    /// error in strict encoded data in LNP message
```
dr-orlovsky, line 44: This is LMP: Lightning messaging protocol :)

dr-orlovsky (approved):

PR #145, opened 2020-11-12
Expose rgb::stash::ExtensionData

zoedberg:
in order to expose `validate` (as requested in LNP-BP/rgb-node#97) the `ExtensionData` type must be exposed (used to create the `Consignment` object)

### DIFF
--- a/src/rgb/mod.rs
+++ b/src/rgb/mod.rs
@@ -36,8 +36,8 @@ pub mod prelude {
         SimplicityScript, TransitionAbi, TransitionAction,
     };
     pub use stash::{
-        Anchor, AnchorId, Consignment, ConsignmentEndpoints, Disclosure, Stash,
-        TransitionData, PSBT_OUT_PUBKEY, PSBT_OUT_TWEAK,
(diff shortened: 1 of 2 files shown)


rajarshimaitra:
Changeset understood. 

It would be better if you make the full PR corresponding to the [issue,](https://github.com/LNP-BP/rgb-node/issues/97) instead of only this small change, so the changes can be reviewed in context. But it seems required, though I haven't tried this on my end. Would be able to review better if I see the full PR.

zoedberg:
I'm still working on the full PR, I will open that when it will be ready for review. But the PR will be opened on the rgb-node, not here, this is the reason I've already opened this.

rajarshimaitra:
Understood.. UtACK..

zoedberg:
Sorry, I've just realized that this PR is not necessary. Initially I made `validate` getting the separated parameters of the `Consignment`, but then I've changed the code on rgb-node to directly get the `Consignment` object. So if that's correct, we don't need to merge this PR.

dr-orlovsky (approved):
While `validate` works with the full consignment, this exposure is still useful

PR #146, opened 2020-11-16
Add serde (de)serialization of Consignment

zoedberg:
this PR is needed in order to do the task

> Add validate and accept functions to support acceptance of transfers

defined in LNP-BP/rgb-sdk#2


dr-orlovsky:
https://github.com/LNP-BP/rgb-sdk/pull/21#issuecomment-727873107

dr-orlovsky (commented):
For historical records: while the implementation is correct, consignment serialization as string is a very bad idea: the file can be of many tens of megabytes in binary form and will grow x10 with JSON/YAML. That's why serde serialization is not needed and it must be passed as a binary data using strict encoding format.

PR #149, opened 2020-11-19
add publish testing workflow

dieeasy:
this PR adds a "Publish" workflow to catch early issues that may arise at publishing time

as briefly discussed, if this introduces a significant slowdown in day to day operations, we can wait to merge this or devise a new scheme to handle code contributions (and contributors) that require the least friction

notes:
- `cargo publish` is called with the `--dry-run` option to skip the real publishing
- a first fix that's been found by the workflow is included (`error: all dependencies must have a version specified when publishing.`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,7 @@ zmq = { version = "~0.9", features = ["vendored"] }
 zeromq-src = { version = "~0.1", git = "https://github.com/LNP-BP/zeromq-src-rs", branch = "fix/cmake" }
 
 [target.'cfg(target_os="windows")'.dependencies]
-openssl = { features = ["vendored"] }
+openssl = { version = "~0.10", features = ["vendored"] }
 
 [patch.crates-io]
 # required to build on 32-bit architectures
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,7 +91,7 @@ torut = { version = "~0.1.6", features = ["v2", "v3"] }
 zmq = { version = "~0.9", features = ["vendored"] }
 
 [target.'cfg(target_os="ios")'.dependencies]
-zeromq-src = { git = "https://github.com/LNP-BP/zeromq-src-rs", branch = "fix/cmake" }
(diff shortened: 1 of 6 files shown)


dieeasy:
android builds were failing due to `cc` and `socket2`
fixed versions for those two packages to working ones and it now builds successfully

Cargo.toml
```diff
@@ -88,13 +88,15 @@ async-trait = { version = "~0.1.30", optional = true }
 torut = { version = "~0.1.6", features = ["v2", "v3"] }
 
 [target.'cfg(target_os="android")'.dependencies]
+cc = { version = "=1.0.41" }
+socket2 = { version = "=0.3.15" }
 zmq = { version = "~0.9", features = ["vendored"] }
 
 [target.'cfg(target_os="ios")'.dependencies]
-zeromq-src = { git = "https://github.com/LNP-BP/zeromq-src-rs", branch = "fix/cmake" }
+zeromq-src = { version = "~0.1", git = "https://github.com/LNP-BP/zeromq-src-rs", branch = "fix/cmake" }
```
dr-orlovsky, line 96: Why do we need the version part? I thought it is incompatible/not working when you also provide git and branch

dr-orlovsky:
A question: is this comparable with https://github.com/LNP-BP/rust-lnpbp/pull/143? I mean if I merge both, would be there no conflicts between them?

Cargo.toml
dieeasy: that's something this pre-publishing check uncovered
without version specification, `cargo publish --dry-run` returns the following error:
```
error: all dependencies must have a version specified when publishing.
dependency `zeromq-src` does not specify a version
Note: The published dependency will use the version from crates.io,
the `git` specification will be removed from the dependency declaration.
```

dr-orlovsky (approved):

dieeasy:
> A question: is this comparable with #143? I mean if I merge both, would be there no conflicts between them?

they're different:
- this PR introduces a pre-publishing check that helps keep the package in a publish-able state
- #143 aims to actually publish the package

I used a name that's not specific enough for this one, so I'd rename it to `pre-publish`, leaving the current name available for #143

PR #150, opened 2020-11-20
Implement Autoconceal for other collection type

### LINKED ISSUE #91: Add generic implementation of AutoConceal trait for collection types
Here: https://github.com/LNP-BP/rust-lnpbp/blob/6010402f8e49fd05ad23f8c0e93ffe079d33d378/src/rgb/contract/conceal.rs#L26

rajarshimaitra:
Fixes #91 
It was a good first issue but it was dangling for a few months now, and I thought it can be useful.

### DIFF
--- a/src/rgb/contract/conceal.rs
+++ b/src/rgb/contract/conceal.rs
@@ -12,6 +12,7 @@
 // If not, see <https://opensource.org/licenses/MIT>.
 
 use core::cmp::Ord;
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 
 use super::seal;
 
@@ -22,21 +23,72 @@ pub trait AutoConceal {
     fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize;
 }
 
-// TODO: Do an auto implementation for and other collection types
-
 impl<T> AutoConceal for Vec<T>
+where
+    T: AutoConceal,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let count = self
+            .iter_mut()
+            .fold(0usize, |sum, item| sum + item.conceal_except(seals));
+        count
+    }
+}
+
+impl<T> AutoConceal for BTreeSet<T>
 where
     T: AutoConceal + Ord + Clone,
 {
     fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
         let mut count = 0;
-        let mut new_self = Vec::<T>::new();
+        let mut new_self = BTreeSet::<T>::new();
         for item in self.iter() {
             let mut new_item = item.clone();
             count += new_item.conceal_except(seals);
-            new_self.push(new_item);
+            new_self.insert(new_item);
         }
         *self = new_self;
+        count
+    }
+}
+
+impl<K, V> AutoConceal for BTreeMap<K, V>
+where
+    V: AutoConceal,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let count = self
+            .iter_mut()
+            .fold(0usize, |sum, item| sum + item.1.conceal_except(seals));
+        count
+    }
+}
+
+impl<T> AutoConceal for HashSet<T>
+where
+    T: AutoConceal + Ord + Clone + std::hash::Hash,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let mut count = 0;
+        let mut new_self = HashSet::<T>::new();
+        for item in self.iter() {
+            let mut new_item = item.clone();
+            count += new_item.conceal_except(seals);
+            new_self.insert(new_item);
+        }
+        *self = new_self;
+        count
+    }
+}
+
+impl<K, V> AutoConceal for HashMap<K, V>
+where
+    V: AutoConceal,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let count = self
+            .iter_mut()
+            .fold(0usize, |sum, item| sum + item.1.conceal_except(seals));
         count
     }
 }
--- a/src/rgb/contract/assignments.rs
+++ b/src/rgb/contract/assignments.rs
@@ -1900,7 +1900,7 @@ mod test {
         // Conceal all without any exception
         // This will create 2 Confidential and 2 ConfidentialState type
         // Assignments
-        hash_type.conceal_except(&Vec::<seal::Confidential>::new());
+        assert_eq!(2, hash_type.conceal_all());
 
         // Precomputed values of revealed seals in 2 ConfidentialState type
         // Assignments


src/rgb/contract/conceal.rs
```diff
@@ -40,3 +39,73 @@ where
...
+}
+
+impl<T> AutoConceal for HashSet<T>
+where
+    T: AutoConceal + Ord + Clone + std::hash::Hash,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let mut count = 0;
+        let mut new_self = HashSet::<T>::new();
+        for item in self.iter() {
```
dr-orlovsky, line 85: Here we can use `iter_mut` and avoid multiple allocations

src/rgb/contract/conceal.rs
```diff
@@ -40,3 +39,73 @@ where
...
+
+impl<K, V> AutoConceal for HashMap<K, V>
+where
+    V: AutoConceal + Ord + Clone + std::hash::Hash,
+    K: Ord + Clone + std::hash::Hash,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let mut count = 0;
+        let mut new_self = HashMap::new();
+        for (k, v) in self.iter() {
```
dr-orlovsky, line 103: Ibid

dr-orlovsky (changes requested):
While BTree* collection types do not provide mutable iterators or `with_capacity`, they are provided by Hash* collections, so I propose to use them and avoid allocations

rajarshimaitra:
On the contrary, I found `iter_mut` for `Vec`, `BtreeMap` and `HashMap`.

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,77 @@ pub trait AutoConceal {
...
 where
     T: AutoConceal + Ord + Clone,
 {
     fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
         let mut count = 0;
-        let mut new_self = Vec::<T>::new();
+        for item in self.iter_mut() {
+            count += item.conceal_except(seals);
+        }
+        count
```
dr-orlovsky, line 35: Nit, however more idiomatic implementation for the whole function in a single line

```suggestion
        self.iter_mut().fold(0usize, |sum, &mut item| sum + item_conceal_except(seals))
```

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,77 @@ pub trait AutoConceal {
...
+    }
+}
+
+impl<T> AutoConceal for BTreeSet<T>
+where
+    T: AutoConceal + Ord + Clone,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let mut count = 0;
+        let mut new_self = BTreeSet::<T>::new();
```
dr-orlovsky, line 45: Not sure whether `BTreeSet` (or `HashSet`) provides `with_capacity()`, but if it does this is important to use it, otherwise we may get a re-allocation per each function cycle.

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,77 @@ pub trait AutoConceal {
...
+where
+    V: AutoConceal + Ord + Clone,
+    K: Ord + Clone,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let mut count = 0;
+        for (_, v) in self.iter_mut() {
+            count += v.conceal_except(seals);
+        }
+        count
```
dr-orlovsky, line 66: See sample with iterators above

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,77 @@ pub trait AutoConceal {
...
+    }
+}
+
+impl<T> AutoConceal for HashSet<T>
+where
+    T: AutoConceal + Ord + Clone + std::hash::Hash,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let mut count = 0;
+        let mut new_self = HashSet::<T>::new();
```
dr-orlovsky, line 76: Ibid regarding `with_capacity`

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,77 @@ pub trait AutoConceal {
...
+where
+    V: AutoConceal + Ord + Clone + std::hash::Hash,
+    K: Ord + Clone + std::hash::Hash,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let mut count = 0;
+        for (_, v) in self.iter_mut() {
+            count += v.conceal_except(seals);
+        }
+        count
```
dr-orlovsky, line 97: Ibid regarding single-line iterator

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,77 @@ pub trait AutoConceal {
...
+            new_self.insert(new_item);
         }
         *self = new_self;
         count
     }
 }
+
+impl<K, V> AutoConceal for HashMap<K, V>
+where
+    V: AutoConceal + Ord + Clone + std::hash::Hash,
```
dr-orlovsky, line 89: Not sure why we need value here to implement `Hash` (it's only required for keys). Also `Ord` is unnecessary for both key and value in `HashMap` and values of `HashSet` (this is needed only for keys in `BTreeMap` and values of `BTreeSet`)

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,77 @@ pub trait AutoConceal {
...
+        for (_, v) in self.iter_mut() {
+            count += v.conceal_except(seals);
+        }
+        count
+    }
+}
+
+impl<T> AutoConceal for HashSet<T>
+where
+    T: AutoConceal + Ord + Clone + std::hash::Hash,
```
dr-orlovsky, line 72: `Ord` is unneeded (see below).

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,77 @@ pub trait AutoConceal {
...
         }
         *self = new_self;
         count
     }
 }
+
+impl<K, V> AutoConceal for HashMap<K, V>
+where
+    V: AutoConceal + Ord + Clone + std::hash::Hash,
+    K: Ord + Clone + std::hash::Hash,
```
dr-orlovsky, line 90: No need for `Clone` here and everywhere where we use mutating iterators

dr-orlovsky (changes requested):
Just a few more improvements

dr-orlovsky:
The CI failure should gone with your next push (I merged fixing PR)

src/rgb/contract/conceal.rs
rajarshimaitra: Nice!! Used it.

src/rgb/contract/conceal.rs
rajarshimaitra: It seems both the sets don't provide `with_capacity()` so we are stuck with this. :(

src/rgb/contract/conceal.rs
rajarshimaitra: Done

src/rgb/contract/conceal.rs
rajarshimaitra: See above..

src/rgb/contract/conceal.rs
rajarshimaitra: Done..

src/rgb/contract/conceal.rs
rajarshimaitra: `Ord` and `Hash` traits are required because without them `insert()` doesn't work.

src/rgb/contract/conceal.rs
rajarshimaitra: See above..

src/rgb/contract/conceal.rs
rajarshimaitra: Done..

rajarshimaitra:
Updated as per comments, some remains. Added a small change in assignment test to check for autoconceal return value.

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,66 @@ pub trait AutoConceal {
...
 
-// TODO: Do an auto implementation for and other collection types
-
 impl<T> AutoConceal for Vec<T>
+where
+    T: AutoConceal,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        self.iter_mut()
+            .fold(0usize, |_, item| item.conceal_except(seals))
```
dr-orlovsky, line 32: This will not return a sum! It will return only the number of the concealed items in the last call. Pls see the code that I wrote in last review which is correct. Why you removed the first closure arg?

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,66 @@ pub trait AutoConceal {
...
+    }
+}
+
+impl<K, V> AutoConceal for BTreeMap<K, V>
+where
+    V: AutoConceal,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        self.iter_mut()
+            .fold(0usize, |_, item| item.1.conceal_except(seals))
```
dr-orlovsky, line 59: Ibid

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,66 @@ pub trait AutoConceal {
...
     }
 }
+
+impl<K, V> AutoConceal for HashMap<K, V>
+where
+    V: AutoConceal,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        self.iter_mut()
+            .fold(0usize, |_, item| item.1.conceal_except(seals))
```
dr-orlovsky, line 86: Ibid

dr-orlovsky (changes requested):
One mistake with iterators

dr-orlovsky:
And CI seems to require one more rebase to include just merged CI PR

src/rgb/contract/conceal.rs
rajarshimaitra: Because as per the definition of fold here https://doc.rust-lang.org/std/iter/trait.Iterator.html I thought it will auto accumulate over the whole iteration. I didn't need to store it and the first argument was giving `unused variable` warnings so I removed it.  You are right, it does only return the last value. Changing it.

rajarshimaitra:
Updated and rebased.

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,72 @@ pub trait AutoConceal {
...
+
+impl<K, V> AutoConceal for HashMap<K, V>
+where
+    V: AutoConceal,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let count = self
+            .iter_mut()
+            .fold(0usize, |sum, item| sum + item.1.conceal_except(seals));
+        count
```
dr-orlovsky, line 92: Sorry for niting, but this can be further simplified to

```suggestion
        self
            .iter_mut()
            .fold(0usize, |sum, item| sum + item.1.conceal_except(seals))
```

src/rgb/contract/conceal.rs
```diff
@@ -22,21 +23,72 @@ pub trait AutoConceal {
...
+
+impl<K, V> AutoConceal for BTreeMap<K, V>
+where
+    V: AutoConceal,
+{
+    fn conceal_except(&mut self, seals: &Vec<seal::Confidential>) -> usize {
+        let count = self
+            .iter_mut()
+            .fold(0usize, |sum, item| sum + item.1.conceal_except(seals));
+        count
```
dr-orlovsky, line 63:
```suggestion
        self
            .iter_mut()
            .fold(0usize, |sum, item| sum + item.1.conceal_except(seals))
```

dr-orlovsky (approved):
While there is another possible tiny improvement merging now and stopping from niting

src/rgb/contract/conceal.rs
rajarshimaitra: Oh I can return it without holding? Neat. Doing this in a small PR next.

PR #154, opened 2020-11-30
Moving onto master-branch dependency for all subcrates

dr-orlovsky:


rajarshimaitra:
LGTM. I was about to open the PR. This will work too. I will check in my local once this is merged..

rajarshimaitra:
Regarding the build failures in CI. I was seeing the same failures in rgb-node while trying to build from lnpbp-master. It's weird that similar failures are occuring here..

dr-orlovsky:
Seems like `serde_with` got broken few days ago with new patch/minor version. Local builds succeed only b/c of older cache.

Now need to figure out which version will still work...

dr-orlovsky:
Funny enough the same version was building just three days ago...

rajarshimaitra:
Rust is weird. Sometimes it seems to have its own mind.

dr-orlovsky:
I have found the reason: it's `serde_with_macros` crate which was updated from 1.2.2. to 1.3.0. We do not use it directly, but it's used by serde_with, which allows minor updates and this update is a buggy one. So each time you do cargo update on master you will get the same error

dr-orlovsky:
Pushed fix to the serde problem, NB to everyone - now we need to use the master dependency

zoedberg:
> NB to everyone - now we need to use the master dependency

Sorry, what do you mean here? The `master` dependency of what and where?

Moreover, are you sure this fix is needed? Is this related to the `serde` issues we discussed on IRC this morning? I don't have any problems by building rust-lnpbp from `master` and I don't see any problems on the CI either

UkolovaOlga:
@zoedberg >Moreover, are you sure this fix is needed? Is this related to the serde issues we discussed on IRC this morning? 

yes, that was the same issue from the IRC chat

dr-orlovsky:
@zoedberg just try `cargo update && cargo build --all-features` and you will see the problems which can't be solved w/o this fix. Notified you just if you will witness this in other projects (RGB Node, RGB SDK) you know what to do. And yes, if you do `cargo update` before building you would not be able to build RGB Node and SDK as well

zoedberg:
> just try cargo update && cargo build --all-features

aaah ok, now it all makes sense to me. I didn't understand you had to update dependencies (on IRC you mentioned you also tried to update deps, but I thought that was just a test to fix the issue you were experiencing). If updating deps is a necessity for this project, let's merge this, otherwise we can wait until it's really needed.

for now I don't think I need to update deps on the sdk, and I don't think that rgb-node needs it either, so I won't need to switch to lnpbp `master` if that's what you were meaning (moreover, we can use tags or fixed branches if we want to avoid releasing too often. using `master` can be very dangerous)

dr-orlovsky:
@zoedberg that's not working like "you can simply do not update dependencies" :)

If you need to connect RGB Node or RGB SDK to your project today, you will have beta-2 failing to build and won't be able to solve that issue w/o this fix. That's why I notified you b/c otherwise we will have 100% of devs trying to integrate with RGB creating issues about this fact everywhere.

zoedberg:
> that's not working like "you can simply do not update dependencies" :)

actually, it is ;)

> If you need to connect RGB Node or RGB SDK to your project today, you will have beta-2 failing to build and won't be able to solve that issue w/o this fix. That's why I notified you b/c otherwise we will have 100% of devs trying to integrate with RGB creating issues about this fact everywhere.

Sorry, but I have to contradict you. Beta-2 is not failing to build and the reasons are:
- cargo doesn't use `Cargo.lock` when getting the dependencies of dependencies, but uses the `Cargo.toml`
- beta-2 has a `Cargo.toml` which is different from the one that is now in `master`

dr-orlovsky:
@zoedberg you may try to build https://github.com/pandoracore/bitcoin-pro against `v0.2.0-beta.2` and witness the problem yourself. I spent a day figuring out what's happens.

zoedberg:
> you may try to build https://github.com/pandoracore/bitcoin-pro against rgb-node v0.2.0-beta.2 and witness the problem yourself. I spent a day figuring out what happens.

wait, why are we talking about rgb-node `v0.2.0-beta.2` now? I thought we were talking about rust-lnpbp `v0.2.0-beta.2`, which is the version currently used by rgb-node `v0.2.0-beta.3`, which is the version I'm using on rgb-sdk (both mentioned versions are building correctly, even after the merge of this PR)

dr-orlovsky:
I am demonstrating the problem
1) If you start a new project you will have the same problem with rust-lnpbp b/c there is no Cargo.lock freezing the working version of serde_with_macro
2) you can't fix RGB Node w/o fixing rust-lnpbp b/c fix will require to update Cargo.lock in RGB Node anyway

That's why it is not "you can simply do not update dependencies"

zoedberg:
Now I understand what you say and agree, on new projects it will not work

Since the sdk has the `Cargo.lock` committed, its build still works and there should not be any need to switch to rust-lnpbp `master`, we could wait for its next release.
But this is not true for rgb-node, a user may try to use `install` (even if on the README we suggest using `build`) and therefore experience the problems you're talking about. So your idea of fixing this ASAP is indeed good (I'm sorry we wasted some time on this, I need to think more before answering)

I'm still convinced that using `master` is dangerous, but I've just seen you add a git tag, so I hope everybody will use that

dr-orlovsky:
The master thing is unrelated from the serde fix; I just combined them two in the same PR.

We need to stick to the master internally (for services and derive) between published versions since otherwise the projects linking to LNP/BP master (during development stage) will link against both master _and_ released version, which breaks things (reported by @rajarshimaitra )

PR #156, opened 2020-12-02
Fixing tokio dependency breaking change

dr-orlovsky:
One of tokio dependencies has updated it's dependencies in minor version and led to Cargo.lock conflict with a patch required for android version. I have temporarily disabled it since it's impossible to update upstream crates (adding any dependency there requires update to the Cargo.lock, which fails due to this problem).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -780,7 +780,7 @@ dependencies = [
  "inflate",
  "lazy_static",
  "lightning-invoice",
- "lnpbp_derive 0.2.0-beta.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


dr-orlovsky:
This think broke everything: miow v0.3.6. They have yanked all previous versions and introduced breaking changes with new one: https://crates.io/crates/miow/versions

dr-orlovsky:
Removing socket2 pinned version breaks android build, but anyway newly created projects will not be able to use LNP/BP Core library w/o this at all, so better to have android build broken for now (and fix it later) than being unable to plug in LNP/BP core lib into new projects at all

PR #163, opened 2020-12-08
[WIP] Implementing HTLC extension modification to TxGraph

rajarshimaitra:
This is a WIP on HTLC extension modification to the TXGraph structure. 

This is preliminary and for approach check. There are many rough edges. Instead of describing the issues here, I have added comments in the codebase referring to which discussions can be held here. 

There are many data that are required for constructing these HTLC transactions which are not there in the HTLC structure. I suppose the extension would then need to have access to global channel state variables in order to get that information. Suggestions over this would be helpful.

This might be also a good time to decide how these should be stored on the graph map inside TxGraph.
 
Note: for anchor outputs, we would need clear segregation Anchors, HTLCs, and payment outputs of a commitment Tx. There will be specific modifications on specific outputs. We should think of creating the TxGraph keeping that in mind.

### DIFF
--- a/derive/src/lnp_api.rs
+++ b/derive/src/lnp_api.rs
@@ -272,15 +272,9 @@ impl EncodingSrategy {
 
     pub fn encode_fn(&self) -> TokenStream2 {
         match self {
-            Self::Strict => {
-                quote!(strict_encode)
-            }
-            Self::Bitcoin => {
-                quote!(consensus_encode)
-            }
-            Self::Lightning => {
-                quote!(lightning_encode)
-            }
+            Self::Strict => quote!(strict_encode),
+            Self::Bitcoin => quote!(consensus_encode),
+            Self::Lightning => quote!(lightning_encode),
         }
     }
 
--- a/src/bp/dbc/txout.rs
+++ b/src/bp/dbc/txout.rs
@@ -87,9 +87,7 @@ impl Container for TxoutContainer {
 }
 
 /// [bitcoin::TxOut] containing LNPBP-2 commitment
-#[derive(
-    Wrapper, Clone, PartialEq, Eq, Hash, Default, Debug, Display, From,
-)]
+#[derive(Wrapper, Clone, PartialEq, Eq, Hash, Default, Debug, Display, From)]
 #[display(Debug)]
 pub struct TxoutCommitment(TxOut);
 
--- a/src/bp/scripts/types.rs
+++ b/src/bp/scripts/types.rs
@@ -110,10 +110,10 @@ use bitcoin::{
 use crate::bp::{Chain, CompactDescriptor};
 use crate::strict_encoding;
 
-/// Script which knowledge is required for spending some specific transaction
+/// Script whose knowledge is required for spending some specific transaction
 /// output. This is the deepest nested version of Bitcoin script containing no
 /// hashes of other scripts, including P2SH redeemScript hashes or
-/// witnessProgram (hash or witness script), or public keys
+/// witnessProgram (hash or witness script), or public key hashes
 #[derive(
     Wrapper,
     Clone,
--- a/src/lnp/application/channel.rs
+++ b/src/lnp/application/channel.rs
@@ -205,7 +205,7 @@ pub struct TxGraph {
     pub cmt_locktime: u32,
     pub cmt_sequence: u32,
     pub cmt_outs: Vec<TxOut>,
-    graph: BTreeMap<u16, BTreeMap<u64, Psbt>>,
+    pub graph: BTreeMap<u16, BTreeMap<u64, Psbt>>,
 }
 
 impl TxGraph {
--- a/src/lnp/application/channel.rs
+++ b/src/lnp/application/channel.rs
@@ -205,7 +205,7 @@ pub struct TxGraph {
     pub cmt_locktime: u32,
     pub cmt_sequence: u32,
     pub cmt_outs: Vec<TxOut>,
-    pub graph: BTreeMap<u16, BTreeMap<u64, Psbt>>,
+    graph: BTreeMap<u16, BTreeMap<u64, Psbt>>,
 }
 
 impl TxGraph {
@@ -227,6 +227,22 @@ impl TxGraph {
         self.graph
             .get_mut(&role.into())
             .and_then(|v| v.get_mut(&index.into()))
+    }
+
+    pub fn insert_tx<R, I>(
+        &mut self,
+        role: R,
+        index: I,
+        psbt: Psbt,
+    ) -> Option<Psbt>
+    where
+        R: TxRole,
+        I: TxIndex,
+    {
+        self.graph
+            .entry(role.into())
+            .or_insert(empty!())
+            .insert(index.into(), psbt)
     }
 
     pub fn len(&self) -> usize {
--- a/src/lnp/application/channel.rs
+++ b/src/lnp/application/channel.rs
@@ -40,6 +40,9 @@ use super::Messages;
 pub enum Error {
     /// Extension-specific error: {0}
     Extension(String),
+
+    // HTLC Extension Errors
+    HTLC(String),
 }
 
 /// Marker trait for any data that can be used as a part of the channel state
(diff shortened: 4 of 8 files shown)


dr-orlovsky (approved):
So far it looks to be correct

dr-orlovsky:
Nb: Roles must be global for all extensions, which allow them to understand data of each other; indexes are not.
Also, we should avoid accessing `graph` directly since then we bypass type/index nomenclature. 

I made my commit with the fixes: https://github.com/LNP-BP/rust-lnpbp/pull/163/commits/05169ffced54090b088680c881842fb2d648cd70

rajarshimaitra:
Ok. Thanks for the update. Ya, that is a better way of accessing the map. There is still one issue here. 

- The way transaction are added right now, uses `index` for a single set of `offered`/`recieved` htlcs. So for the next set of htlcs (probably created by the update_htlc message) this index will again start at 0. I think thats not desirable. What we need is to have unique index for each txs. otherwise there will be key collision in the map. May be we read the last index already in the map and start from there? Looking for suggestions.

apart from the map. The following is the next TODO:

-  Get the various channel specific `keys` for htlc construction.

For this I think we need a channel structure, which should have all the channel-specific data required for these constructions. Right now there are two structures that holds these data:
https://github.com/LNP-BP/rust-lnpbp/blob/03f388811d7194eb8143ae5df915b84fa4500c92/src/lnp/application/payment/channel.rs#L67-L79
and
https://github.com/LNP-BP/rust-lnpbp/blob/03f388811d7194eb8143ae5df915b84fa4500c92/src/lnp/application/payment/channel.rs#L169-L178


The Htlc extension should be able to somehow access these data and use them while applying over the TxGraph. I am not quite sure whats the best way to do this. Should i create a simple `Channel` structure and pass it to `Htlc` while its constructed? Looking for suggestions.

Also, exploring for the same I found another `keyset` structure inside `Bolt3` extension here
https://github.com/LNP-BP/rust-lnpbp/blob/03f388811d7194eb8143ae5df915b84fa4500c92/src/lnp/application/payment/constructors/bolt3.rs#L28-L32

This should also follow the global Channel structure data. And I think a keyset here is redundant.

My Initial thought is to have a simple map of`Map<Channel_Id, Channel>` which will be appended for any new channel opening. All other extensions should refer to this map and extract the required data. Let me know if some other philosophy for the same is already implemented and I am missing.

dr-orlovsky:
Can you pls merge from master (or rebase) and resolve conflicts after my recent large-scale refactoring of encodings (strict & lightning)?

dr-orlovsky:
Answering your points: we need to use HTLC extension state (which is channel-specific, i.e. an extension instance is always used by a single and only one channel) to store both tx indexing info and keys. The keys can be read from messages (see message processing in https://github.com/LNP-BP/rust-lnpbp/pull/163/files#diff-94120d5e3bb676a87fa56c19f489a40328e6c58b688febab755a3eae938f29caL58-L73 where you need to fill in mesage handlers extracting key info from LN messages)

rajarshimaitra:
Rebased on top of master.

rajarshimaitra:
The link you pasted points to the diff of this commit. Not sure if thats what you intended. 

So far iterating my understanding. 

- Each extension will have a channel specific state which will be created at the time of its first construction. These data will be constant over incoming messages. Data like keys, and parameters for that channel.
- The extension will also hold Message specific data, lets call it "variable state", like htlc details that needs to be updated in the next call.
-  Each messsage like `Update_Htlc` will regenrate this "variable state". While the channel specfic state will remain unchanged.  

You probably ment this by message processing? https://github.com/LNP-BP/rust-lnpbp/blob/aec2c283c62bfebc4164dfcc28fce5780f40a8c9/src/lnp/application/payment/extenders/htlc.rs#L58-L73

If yes then the rest is straight forward. I dont see any current arrangement inside `Htlc` structure to hold those two types of states, but that can be extended pretty easily. 

Let me know if my understanding is correct.

dr-orlovsky:
> Each extension will have a channel specific state which will be created at the time of its first construction. These data will be constant over incoming messages. Data like keys, and parameters for that channel.

Not sure whether keys are constant and are not updated from the messages in case of HTLC

> The extension will also hold Message specific data, lets call it "variable state", like htlc details that needs to be updated in the next call.
> Each messsage like Update_Htlc will regenrate this "variable state". While the channel specfic state will remain unchanged.

I would not call it "message specific data", I will call it "channel-specific extension state". Yes, messages are processed in the `update_from_peer` function and this state is changed.
This state is also initialized by extension at the very beginning and in most cases this state will be initialized empty (like in HTLC it will be an empty lists of known HTLC secrets and hashes, plus an undefined list of keys (or dumb keys)

> You probably ment this by message processing? 

Yes

rajarshimaitra:
>Not sure whether keys are constant and are not updated from the messages in case of HTLC

There will be two types of data required here. 

One, that will remain common for the channel. Like `to_self_delay`, `dust_limit_satoshis` etc. These will be used for both creating the HTLC outputs as well as handling the logics of weather HTLC will be created or not (not implemented currently). 

Second, commitment specific data. Like `revocation_pubkey`, `local_htlc_pubkey`, `remote_htlc_pukey` etc. These will be used to construct HTLC output as well as HTLC transactions (success/timeout). And they will change for each new commitment transaction.

Question is, how these data will be communicated to the HTLC extension?? I don't see this being done in any form in other extensions also.

The same issue is applicable to `Bolt3` extension. Right now its taking some `keyset` which contains `*_basepiont`s and not implementing any per-commitment logic. Keys will be different for each new commitment tx. But these same keys will be then used in HTLC extension, so they have to be communicated somehow.

>I would not call it "message specific data", I will call it "channel-specific extension state".

These two are diffrent things as explained above. 

>This state is also initialized by extension at the very beginning and in most cases this state will be initialized empty (like in HTLC it will be an empty lists of known HTLC secrets and hashes, plus an undefined list of keys (or dumb keys)

As mentioned above these states are not present in HTLC and Bolt3 extension right now. You can see I have used `dummy` placeholder for these values in the PR. There has to be some way in which these values are communicated to the extension 
 (or the extension calculates them on fly)  when `apply()` will be called over the `TxGraph`.

The next step is to implement these states and I am looking for suggestions on how they should be done.

I can see two ways of doing this, and this can be common for all the extensions.

- Add a channel specific and commitment specific data to individual extensions. The channel specific part will be filled when the extension is first created by a channelmanager (or whoever creates them). The commitment specific part will be updated for each new message.

- Add only message specific part to extensions and allow them to access some global channel-specific state, from where channel-specific data can be read.

dr-orlovsky:
> Question is, how these data will be communicated to the HTLC extension?? I don't see this being done in any form in other extensions also.

You need to maintain the copy of those data in the extension not relying on external sources of truth. You will reconstruct them from the same messages that communicate them to the remote peer

Why we should use such way? In this way we do not need to update channel API on each BOLT change, which happens frequently and we use loose coupling between extensions and main channel model.

It could be that we will need to add another `update_from_local` processing all messages that are sent by the channel to the remote peer - feel free to add it to the extension trait API and add handles to it.

rajarshimaitra:
Ya, probably I am getting somewhere very confused. I am not seeing where and how this data is being handled. 

> You need to maintain the copy of those data in the extension not relying on external sources of truth

Is this already done somewhere? Can you point me to it? 

> You will reconstruct them from the same messages that communicate them to the remote peer.

Massive confusion here. Not every data needed to make a commitment Tx is sent via remote messages. Where will the rest come from?

I guess to totally boil down my question into a single sentence is 

What do I write here for these variables? 

https://github.com/LNP-BP/rust-lnpbp/blob/3ca19c0c446a05da2494daa3cf3c57a2f83887a9/src/lnp/application/payment/extenders/htlc.rs#L92-L99

rajarshimaitra:
Or you are saying it's not at all implemented right now, then I can start extending the extensions to hold those data.

dr-orlovsky:
It's not implemented right now. The channel extensions from LNP/BP Core lib are not used by v0.1 of LNP Node, which manages tx internally. So first we will finish them here and then use from LNP Node in v0.2

rajarshimaitra:
Updated with HTLC preliminary message handling. This section is not complete and I am opening this PR to initiate approach discussions. I just wrote whatever made sense to me. It might be completely off from intended direction.

@dr-orlovsky please take a look and let me know if it makes any sense.

dr-orlovsky:
Merged unapproved since I need to split crate and won't be able to merge after. Requires fixes of the wrong parts - pls see "TODO" comments with the https://github.com/LNP-BP/rust-lnpbp/commit/1bb08b5be77691f89bd961dc56935ca29fb69e0c#diff-c8745da12043e8d6ec35313a1e613eeaf76d01d45152d49db36f2b8e19b05dd4R495 commit

PR #178, opened 2021-02-15
Improvement to client_side_validation crate (CommitEncode for vectors)

claudiosdc:
This adds a standard implementation of the CommitEncode trait (with Merklization strategy) for vectors of a designated type. To activate the implementation, one just needs to implement the new VectorCommitEncode trait marker for that type.

The motivation for this improvement was the need to implement the CommitEncode trait for the Assignments structure of the `rgb-core` crate, which essentially is a vector a complex type.

### DIFF
--- a/client_side_validation/src/client_side_validation.rs
+++ b/client_side_validation/src/client_side_validation.rs
@@ -125,6 +125,17 @@ pub mod commit_strategy {
         }
     }
 
+    // Marker trait for automatically implementing the CommitEncode trait
+    //  (with Merklization strategy) for a vector of that given type
+    pub trait VectorCommitEncode {}
+
+    impl<T> CommitEncodeWithStrategy for Vec<T>
+    where
+        T: VectorCommitEncode
+    {
+        type Strategy = commit_strategy::Merklization;
+    }
+
     impl CommitEncodeWithStrategy for usize {
         type Strategy = UsingStrict;
     }


rajarshimaitra:
The Merkelization strategy has a trait bound of `IntoIterator` here https://github.com/LNP-BP/rust-lnpbp/blob/8e5e1a715efba0704b38fe7d2a24ce71e7b2b23f/client_side_validation/src/client_side_validation.rs#L86

This means the strategy can be applied over a `Vec<T>` if `T` has `commit_encode()` defined. 

So not sure if we really need a separate definition here.

claudiosdc:
> The Merkelization strategy has a trait bound of `IntoIterator` here
> 
> https://github.com/LNP-BP/rust-lnpbp/blob/8e5e1a715efba0704b38fe7d2a24ce71e7b2b23f/client_side_validation/src/client_side_validation.rs#L86
> 
> This means the strategy can be applied over a `Vec<T>` if `T` has `commit_encode()` defined.
> 
> So not sure if we really need a separate definition here.

Yes, I am aware of that. In fact, the change that I am proposing is to allow for that [code](https://github.com/LNP-BP/rust-lnpbp/blob/8e5e1a715efba0704b38fe7d2a24ce71e7b2b23f/client_side_validation/src/client_side_validation.rs#L84-L105) to be used with `Vec<T>`. It is very similar to the definition that already exist for `BTreeSet<T>` (shown below).

https://github.com/LNP-BP/rust-lnpbp/blob/8e5e1a715efba0704b38fe7d2a24ce71e7b2b23f/client_side_validation/src/client_side_validation.rs#L193-L195

The diference is that, since there are already several definitions for the implementation of the CommitEncode trait for vectors of different types (shown below),

https://github.com/LNP-BP/rust-lnpbp/blob/8e5e1a715efba0704b38fe7d2a24ce71e7b2b23f/client_side_validation/src/client_side_validation.rs#L164-L175

we can not simply implement it as shown below, since it would trigger a compiler error:

```rust
impl<T> CommitEncodeWithStrategy for Vec<T> {
    type Strategy = commit_strategy::Merklization;
}
```

So the alternative that I have envisioned was to create a trait marker (`VectorCommitEncode`) to distinguish the new implementations from the ones that already exist.

Finally, I just wanted to emphasize that it would not be possible for someone to define a CommitEncode implementation for a vector of a specific type in the same crate where that type is defined. For instance, the code below would trigger a compiler error since neither `CommitEncodeWithStrategy` nor `Vec` is owned by that crate.

```rust
struct MyStruct {}

impl CommitEncodeWithStrategy for Vec<MyStruct> {
    type Strategy = commit_strategy::Merklization;
}
```

So the right way for doing it is to define a standard CommitEncode implementation for vectors of generic types in the `client_side_validation` crate. And that is exactly what I am proposing here.

client_side_validation/src/client_side_validation.rs
```diff
@@ -125,6 +125,17 @@ pub mod commit_strategy {
         }
     }
 
+    // Marker trait for automatically implementing the CommitEncode trait
+    //  (with Merklization strategy) for a vector of that given type
+    pub trait VectorCommitEncode {}
```
dr-orlovsky, line 130: This approach will not work in a well-defined fashion with the `Merklization` strategy from the above: it's not obvious that using merklization strategy will not merkelize while to merkelize you need to impl a separate trait on the type itself.

dr-orlovsky (changes requested):
Concept NACK. Generally, we need to keep in mind, that this code is not RGB specific and can be used outside of the RGB. We can't use more generic libraries fitting very specific current design of the type system inside RGB.

The current approach requires to use new types for all vectors that has to be merkelized, and implementing a merklization strategy for that newtype; which is more transparent and obvious. Also, merklization is not a property of a type, it is a property of the vector, and for the same type some vectors with that type must be merkelized, and some are not. Finally, merklization may not be about vectors only, but also BTreeSet's and other collection types.

client_side_validation/src/client_side_validation.rs
claudiosdc: To be honest, I do not understand the reasons you are giving for not accepting this change, but you are the lead developer and must know better, and I respect that.

The fact is that, the way this library is right now, one cannot apply CommitEncode with the Merklization strategy to a vector of any specific type (different than `u8`, `u16`, `u32`, and `u64`).

dr-orlovsky: My approach is not to try to fix issues case by case, but have a consistent design on API and type system. If we use a strategy-based derivation, we should not introduce an alternative API without clear distinction.

As for the vectors, I do not see why you can't apply CommitEncoding with merklization for any vector by doing a newtype on it (while the underlying type must also support commit encode). This is a clear way how you can avoid very unpleasant side-effects and bugs like I described in all details in by comment above: merklization is not a property of a type, it is a property of the vector, and for the same type some vectors with that type must be merkelized, and some are not.

dr-orlovsky: @claudiosdc I tried some name refactoring + tests in https://github.com/LNP-BP/rust-lnpbp/pull/179. Collection types are working correctly, no new traits are required. Strange, I am not able to tag you or add you as a reviewer (was able to do that like 10 mins ago)

dr-orlovsky:
Closing in favor of merged https://github.com/LNP-BP/rust-lnpbp/pull/179

PR #179, opened 2021-02-18
Improving commit encodings collection support

dr-orlovsky:
This is an alternative to #178 as a more systematic commit encoding refactoring addressing several issues and providing test coverage

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -703,7 +703,6 @@ dependencies = [
 [[package]]
 name = "lnpbp"
 version = "0.4.0-beta"
-source = "git+https://github.com/LNP-BP/rust-lnpbp#0ccced3445dd5ad76f76c436c00596a926aa3cad"
 dependencies = [
  "amplify",
  "amplify_derive",
@@ -727,6 +726,7 @@ dependencies = [
 [[package]]
 name = "lnpbp"
 version = "0.4.0-beta"
+source = "git+https://github.com/LNP-BP/rust-lnpbp#0ccced3445dd5ad76f76c436c00596a926aa3cad"
 dependencies = [
  "amplify",
  "amplify_derive",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,6 +72,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi",
+ "libc",
+ "winapi 0.3.9",
+]
+
+[[package]]
 name = "autocfg"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -162,6 +173,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "63396b8a4b9de3f4fdfb320ab6080762242f66a8ef174c49d8e19b674db4cdbe"
+
+[[package]]
 name = "byte-tools"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -181,9 +198,9 @@ checksum = "e0dcbc35f504eb6fc275a6d20e4ebcda18cf50d40ba6fabff8c711fa16cb3b16"
 
 [[package]]
 name = "cc"
-version = "1.0.66"
+version = "1.0.67"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4c0496836a84f8d0495758516b8621a622beb77c0fed418570e50764093ced48"
+checksum = "e3c69b077ad434294d3ce9f1f6143a2a4b89a8a2d54ef813d85003a4fd1137fd"
 
 [[package]]
 name = "cfg-if"
@@ -225,7 +242,7 @@ dependencies = [
  "num-traits",
  "serde",
  "time",
- "winapi",
+ "winapi 0.3.9",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


dr-orlovsky:
Rebased on a master tip which now has a fixed CI

claudiosdc:
> This is an alternative to #178 as a more systematic commit encoding refactoring addressing several issues and providing test coverage

Sorry, but I do not see how this could be an alternative for PR #178 when the issue that it was trying to address had to do with implementation of `CommitEncode` for vectors (`Vec<T>`) and not collections in general.

To illustrate what I am saying, if you take the `collections()` test function that is included in this PR and replace the statement that declares the `collection` variable so that the collection is now a vector instead of a binary tree map (as shown below),

```rust
        /*let collection: BTreeMap<usize, Item> = bmap! {
            0 => Item(s!("My first case")),
            1 => Item(s!("My second case with a very long string")),
            3 => Item(s!("My third case to make the Merkle tree two layered"))
        };*/
        let collection = vec![
            Item(s!("My first case")),
            Item(s!("My second case with a very long string")),
            Item(s!("My third case to make the Merkle tree two layered")),
        ];
```

and try to (compile and) execute that test function,

```shell
cargo test -p client_side_validation collections
```

you will get the following error:

```shell
error[E0599]: no method named `commit_serialize` found for struct `Vec<Item>` in the current scope
   --> client_side_validation/src/commit_encode.rs:395:35
    |
395 |         println!("{}", collection.commit_serialize().to_hex());
    |                                   ^^^^^^^^^^^^^^^^ method not found in `Vec<Item>`
    | 
   ::: /Users/claudio/.cargo/registry/src/github.com-1ecc6299db9ec823/amplify-3.0.0/src/strategy.rs:86:1
    |
86  | pub struct Holder<T, S>(T, PhantomData<S>);
    | ------------------------------------------- doesn't satisfy `Holder<[Item], _>: commit_encode::CommitEncode`
    | 
   ::: /Users/claudio/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/vec.rs:300:1
    |
300 | pub struct Vec<T> {
    | -----------------
    | |
    | doesn't satisfy `Vec<Item>: commit_encode::CommitEncodeWithStrategy`
    | doesn't satisfy `Vec<Item>: commit_encode::CommitEncode`
    |
    = note: the method `commit_serialize` exists but the following trait bounds were not satisfied:
            `Vec<Item>: commit_encode::CommitEncodeWithStrategy`
            which is required by `Vec<Item>: commit_encode::CommitEncode`
            `[Item]: Sized`
            which is required by `[Item]: commit_encode::CommitEncode`
            `[Item]: commit_encode::CommitEncodeWithStrategy`
            which is required by `[Item]: commit_encode::CommitEncode`
            `[Item]: Clone`
            which is required by `[Item]: commit_encode::CommitEncode`
            `Holder<[Item], _>: commit_encode::CommitEncode`
            which is required by `[Item]: commit_encode::CommitEncode`
```

Showing that the `CommitEncode` trait is not automatically implemented for `Vec<T>` as it is for `BTreeMap<K,V>` and `BTreeSet<T>`.

dr-orlovsky:
> Showing that the CommitEncode trait is not automatically implemented for Vec<T> as it is for BTreeMap<K,V> and BTreeSet<T>.

Correct, and that is done deliberately, as I commented in #178, otherwise we will have an unpredictable behavior. Each time you need `CommitEncode` on a specific vector type, you need to do `impl CommitEncodeWithStrategy for Vec<ConcreteType> { type Strategy = Merklization }` explicitly - no new trait is needed.

Even if you'd like to support arbitrary `Vec`, you need not a new trait but instead implementing generic patter like it's done in the original `commit_encode` code for `BTreeMap`/`BTreeSet`:

```rust
    impl<T> CommitEncodeWithStrategy for Vec<T> {
        type Strategy = Merklization;
    }
```

However I think this is quite risky and instead considering removing this auto implementations for `BTreeMap`/`BTreeSet` and having manual implementations on each concrete type.

claudiosdc:
> Each time you need `CommitEncode` on a specific vector type, you need to do `impl CommitEncodeWithStrategy for Vec<ConcreteType> { type Strategy = Merklization }` explicitly - no new trait is needed.

Yes, that is what I initially thought and tried. However, this is not possible. For instance, if you try to do the following implementation in the `contract::assignments` module of the `rgb-core` crate (see code below),

```rust
impl CommitEncodeWithStrategy for Vec<OwnedState<PedersenStrategy>> {
    type Strategy = commit_strategy::Merklization;
}
```

you will get the following error when you try to compile the code:

```shell
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
  --> src/contract/assignments.rs:51:1
   |
51 | impl CommitEncodeWithStrategy for Vec<OwnedState<PedersenStrategy>> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^---------------------------------
   | |                                 |
   | |                                 `Vec` is not defined in the current crate
   | impl doesn't use only types from inside the current crate
   |
   = note: define and implement a trait or new type instead
```

claudiosdc:
> Even if you'd like to support arbitrary `Vec`s, you need not a new trait but instead implementing generic pattern like it's currently done in the original `commit_encode` code for `BTreeMap`/`BTreeSet` https://github.com/LNP-BP/rust-lnpbp/blob/master/client_side_validation/src/client_side_validation.rs#L190-L195:
> 
> ```rust
>     impl<T> CommitEncodeWithStrategy for Vec<T> {
>         type Strategy = Merklization;
>     }
> ```

Unfortunately, this is also not possible because this generic implementation would conflict with the specific implementations for `Vec<u8>`, `Vec<u16>`, `Vec<u32>`, and `Vec<`u64`>` that already exist. I have also tried it.

dr-orlovsky:
Right, you can't implement foreign traits on the foreign types (like `Vec` or `HashMap`). Thus we will need a generic implementation inside `client_side_verify`. It can be done by removing the existing implementations on different integers (since they are not needed anyway and will be covered by the generic implementation). I do not like such generalization, but afraid no other choice here... Will add a commit in a moment.

dr-orlovsky:
@claudiosdc done with 6b29111188d55a456a85d7f76121cf77756fd432

claudiosdc:
> @claudiosdc done with [6b29111](https://github.com/LNP-BP/rust-lnpbp/commit/6b29111188d55a456a85d7f76121cf77756fd432)

Yes, I guess this will do the trick. Thanks.

I believe we can close PR #178 in favor of this one.

dr-orlovsky:
Thank you for catching that. I do not remember the reason behind that, but I assume I had it when I was doing all that custom implementations (instead of deriving `Vec` the same way I derived `BTreeSet`/`BTreeMap`, especially with this specific exclusion). Will do the investigation tomorrow.

